PR #51, opened 2014-12-05
[TASK] Use PHP 5.5 as minimum requirement

kdambekalns:
Removes the user land definition of json_last_error_msg()


skurfuerst:
As PHP 5.5 is a minimum requirement for Flow 3.0, this is not a problem.

PR #59, opened 2014-12-22
[BUGFIX] Prevent duplicate workspace query filter

aertmann:


kdambekalns:
:+1: by reading

skurfuerst:
absolutely makes sense. Thanks!

PR #62, opened 2014-12-22
[!!!][FEATURE] Make search metadata available

aertmann:
Changes the return format of the execute method to return an array
with a ``nodes`` key containing an array of results. Each result
consists of a ``node`` key with the node instance and a ``meta`` key
containing metadata about the result. Additionally the metadata from
the whole request is available along ``nodes`` as ``response``.

This allows for additional metadata to be passed along, needed for display
of advanced results.

Example of output::

  array(
  	'nodes' => array(
  		array(
  			'node' => NodeInterface,
  			'meta' => array
  		),
  		array(
  			'node' => NodeInterface,
			'meta' => array
  		)
  	),
  	'response' => array
  )

Adjust existing code::

  searchQuery.execute() => searchQuery.execute().nodes

  ${q(node).*} 0> $q(result.node.*}

### DIFF
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
@@ -63,6 +63,13 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
 	protected $from;
 
 	/**
+	 * These fields are not accepted in a count request and must therefore be removed before doing so
+	 *
+	 * @var array
+	 */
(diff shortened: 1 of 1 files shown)


hlubek:
:-1: from me, because that totally breaks all existing usages. We have implemented a feature for pagination using a special QueryResultInterface implementation that could allow to do this much cleaner by having additional getters for the metadata. The code is currently in development but I might be able to share it soon.

kitsunet:
Mmm, it would also diverge from SimpleSearch then. I guess the TYPO3.TYPO3CR.Search package should define a result container that can be extended (with metadata for example)

skurfuerst:
Agreed to hlubek; a custom QueryResultInterface is a good idea I think.

skurfuerst:
Closing, because this has been implemented differently in #68 after disucssion.

PR #63, opened 2014-12-22
[FEATURE] Highlighting support in fulltext search results

aertmann:
Call ``.highlight()`` to wrap search term matches found in fulltext search results. Matches are wrapped in ``<em>`` tags.

### DIFF
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
@@ -63,6 +63,13 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
 	protected $from;
 
 	/**
+	 * These fields are not accepted in a count request and must therefore be removed before doing so
+	 *
+	 * @var array
+	 */
+	protected $unsupportedFieldsInCountRequest = array('fields', 'sort', 'from', 'size');
+
+	/**
 	 * The ElasticSearch request, as it is being built up.
 	 * @var array
 	 */
@@ -419,7 +426,14 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
 	 */
 	public function count() {
 		$timeBefore = microtime(TRUE);
-		$response = $this->elasticSearchClient->getIndex()->request('GET', '/_count', array(), json_encode($this->request));
+		$request = $this->request;
+		foreach ($this->unsupportedFieldsInCountRequest as $field) {
+			if (isset($request[$field])) {
+				unset($request[$field]);
+			}
+		}
+
+		$response = $this->elasticSearchClient->getIndex()->request('GET', '/_count', array(), json_encode($request));
 		$timeAfterwards = microtime(TRUE);
 
 		$treatedContent = $response->getTreatedContent();
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
@@ -67,7 +67,7 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
 	 *
 	 * @var array
 	 */
-	protected $unsupportedFieldsInCountRequest = array('fields', 'sort', 'from', 'size');
+	protected $unsupportedFieldsInCountRequest = array('fields', 'sort', 'from', 'size', '_source', 'highlight');
 
 	/**
 	 * The ElasticSearch request, as it is being built up.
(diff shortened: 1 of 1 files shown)


Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -449,6 +468,28 @@ public function fulltext($searchWord) {
...
 
 	/**
+	 * Wrap search term matches found in the result for highlighting purposes. Only available for fulltext searches.
+	 *
+	 * @param integer $fragmentSize
+	 * @param integer $fragmentCount
+	 * @return QueryBuilderInterface
+	 */
+	public function highlight($fragmentSize = 150, $fragmentCount = 2) {
+		$this->request['_source'] = array('__fulltext');
```
skurfuerst: I do not yet understand why we need to load the _source as well? The "highlight" request works nevertheless, even if this line is removed.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -449,6 +468,28 @@ public function fulltext($searchWord) {
...
+		$this->request['highlight'] = array(
+			'fields' => array(
+				'text' => array(
+					'force_source' => TRUE,
```
skurfuerst: I do not think we need this force_source; it will work nevertheless without.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -449,6 +468,28 @@ public function fulltext($searchWord) {
...
+	 *
+	 * @param integer $fragmentSize
+	 * @param integer $fragmentCount
+	 * @return QueryBuilderInterface
+	 */
+	public function highlight($fragmentSize = 150, $fragmentCount = 2) {
+		$this->request['_source'] = array('__fulltext');
+		$this->request['highlight'] = array(
+			'fields' => array(
+				'text' => array(
```
skurfuerst: We should clearly reference __fulltext.text instead of just text.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -449,6 +468,28 @@ public function fulltext($searchWord) {
 	}
 
 	/**
+	 * Wrap search term matches found in the result for highlighting purposes. Only available for fulltext searches.
```
skurfuerst: can't we just simply *ALWAYS* run highlight requests as soon as a fulltext query is involved? Would keep the API in sync with SimpleSearch.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
aertmann: for us no highlight was done unless the _source parameter was set, it made no sense though.. let's try without it and I'll test it in our environment and see if it works this fine :)

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
aertmann: probably related to the other one, but I'm no elastic search expert and didn't do this so I can't say for sure..

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
skurfuerst: OK; I just tested it with ElasticSearch 1.4.4 and it worked nicely

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
aertmann: don't know the details well enough to chime in here, if you think so and it works then that's fine.. it's a bit too long ago I helped getting it to work..

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
aertmann: don't think so since the output changes because of it and that might not be desired

skurfuerst:
Closing, because this has been re-implemented in #68

PR #66, opened 2015-01-20
[FEATURE] Implement QueryResultInterface in execute()

hlubek:
This change introduces a `ElasticsearchQueryResult` object that implements `QueryResultInterface` and can be used by the default `<f:widget.paginate>` view helper.

The result objects allows to integrate metadata in a compatible way without introducing nested arrays (e.g. `results.nodes`).


skurfuerst:
I like this change and will try to base the changes of Aske upon this one. let's see how far I'll get.

skurfuerst:
Closing, because #68 contains this change and improves them further.

PR #69, opened 2015-03-30
[TASK] Use sub command to limit memory usage during indexing

dfeyer:
This change use a sub command for each workspace indexing to limit the
memory usage during indexing.


dfeyer:
I close this one, we use https://github.com/ttreeagency/Flowpack.ElasticSearch.ContentRepositoryQueueIndexer to support huge amount of nodes.

Issue #70, opened 2015-04-09
ElasticSearch timezone configuration (since version 1.4)


whythecode:
Since version 1.4 there's a configuration for setting the time_zone to be used in a query (See [the docs](https://github.com/elastic/elasticsearch/blob/master/docs/reference/query-dsl/filters/range-filter.asciidoc#date-options), [the commit](https://github.com/elastic/elasticsearch/commit/d28b8c6dae46c922d5f52f0a59d9e1d0b4d37ad7), and [the original issue](https://github.com/elastic/elasticsearch/issues/3729)). (Note that this is apparently still missing for aggregations.)

This is important for all queries which use relative dates and/or "magic" keywords ("now", etc).

One thing that's not clear to me, is how ElasticSearch figures out what "now" means. In either case, it's a lot safer if we make sure it's using the currently configured PHP Timezone, since that's what the node dates also use.

It would probably be ideal to add this option to the default ElasticSearchQueryBuilder query.

yaramayer:
Hi @skurfuerst,

it seems ElasticSearch actually uses Timestamps when creating "now" etc, so as long as the dates are correctly indexed in the first place, everything should be fine. I think we can close this. Can you please verify (insanity check)?

Thanks a lot!

PR #72, opened 2015-04-09
[BUGFIX] Hide Before/After Dates indexed with PHP timezone

yaramayer:
Changed the "_hiddenBeforeDateTime" and "_hiddenAfterDateTime" Search Index Configuration to give over the PHP's currently configured Timezone. That way, considering everything is configured correctly in the server (same locale as PHP), "now" and all other queries should work as expected.

Related: Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor#70

### DIFF
--- a/Configuration/NodeTypes.yaml
+++ b/Configuration/NodeTypes.yaml
@@ -76,7 +76,7 @@
           type: date
           include_in_all: false
           format: 'date_time_no_millis'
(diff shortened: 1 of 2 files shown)


yaramayer:
Oops, updating Doc, just a sec

skurfuerst:
Thanks so much :+1: :)

yaramayer:
np! :) :)

PR #76, opened 2015-06-08
[BUGFIX] After changing NodeType indexes contains two documents

dfeyer:
In ES the ID property is unique inside a single document type. With the
current implementation, after a NodeType change, we create a new document
with the new NodeType, but the old one stay in the index.

This change solve the issue with a DELETE query to check is an other
document share the same _id before indexing new values.


skurfuerst:
Looks good to me!

kdambekalns:
Wow, this broke fulltext indexing - whenever a document node itself was indexed (again), the document in the index was deleted, thus removing all existing `__fulltextParts` and by this removing the child nodes' content from the fulltext index.

The reason: the `str_replace()` on the node type name is plain wrong, it would have been better to use `NodeTypeMappingBuilder::convertNodeTypeNameToMappingName()` - that would have used `-` instead of `/`.

You two owe me and cron IT a few :beers:…

Issue #44, opened 2014-10-09
Multiple Datasets in Index


mgoldbeck:
For one Node there are several datasets in the Elastic Index. So there are to many Nodes indexed which has at least influences for the result number, when all results are shown,
For searching with a dedicated search term the result number is the correct one. (SEARCHING DIRECTLY IN THE INDEX))

This behavior probably depends on the first indexing of the data. For every workspace, all nodes are indexed. If there are no nodes in the user-jondoe workspace the nodes are indexed for workspace live.

mgoldbeck:
As short update. Still the same behaviour and even worse with more than one dimension.

dfeyer:
@mgoldbeck Can you check if your issue is solved by 3605673b5efa392aed8818efefa59f57ae8939e3 and 041b556fd80850f372848327be711f2d2513a0a2 (merged on master)

dfeyer:
@mgoldbeck Can you update this one based on recent master ?

mgoldbeck:
@dfeyer I have no access to this index any longer ;) But will try to replay it with other data.

dfeyer:
@mgoldbeck Any news ?

mgoldbeck:
Will close it because I can't reproduce it.

PR #78, opened 2015-07-16
[BUGFIX] Correctly filter for dimension combinations in query

hlubek:
This change adds a terms filter to the query method that includes the
context dimension values to fetch only matching documents. This works
for most dimension usages except complex use-cases with overlapping
fallback chains.

The result count should now be correct when querying nodes with multiple
variants, so it solves at least one problem of #44.


skurfuerst:
Looks great for me :)

dfeyer:
We have a project with the given dimensions:

```yaml
    contentDimensions:
      'language':
        default: 'fr_FR'
        defaultPreset: 'fr'
        label: 'Language'
        icon: 'icon-language'
        presets:
          'fr':
            label: 'French'
            values: ['fr_FR']
            uriSegment: 'fr'
          'de':
            label: 'German'
            values: ['de_DE', 'fr_FR']
            uriSegment: 'de'
```

Search work nicely in French, but the German search return only nodes translated to german so no fallback support. Did our configuration fall in the "complex use-cases with overlapping
fallback chains." ?

PR #80, opened 2015-07-27
[FEATURE] allow indexing only live workspace

skurfuerst:


aertmann:
Hey, thanks for submitting this one. I'm okay with this change, however it's a workaround to me. The problem is essentially that we haven't found a way to return results across workspace/dimensions yet according to fallback rules. I've looked into a couple of times, but never found a way in ES to group results and choose one by priority. If anybody knows of a way, I'd definitely be interested.

Also it makes the search unusable in the backend if any non-published stuff should be found.

kitsunet:
You need to multiply the content into all possible combinations, there is no other way. So the search doesn't need any fallback logic but can just select from a specific dimension combination. You can reduce the amount of index data again by "tagging" the same data with multiple dimension combinations in the index. That's basically what I do with simple search. IMHO only real solution. The search shouldn'T know anything about the dimension fallback logic.

hlubek:
Having a flat search is much easier and performant. What bugs me is, that the indexing can really take a long time if the multiplication of N nodes * W workspaces * D_1*...*D_n dimension presets gets large. I'd suppose that it starts to get infeasible with a dozen editors and 2 dimensions with a dozen or more presets each.

So having something more intelligent on the indexing side could at least reduce the amount tremendously for sparse variations. With aggregate node types we basically can find all existing node variants below the aggregate and then reverse-calculate the matching combinations of workspaces and dimensions (in theory). That way it's much cheaper and we use way less storage in ES because many combinations of W,D_1,...,D_n can share the same document if node variants for that node (or child-nodes for an aggregate) don't differ.

skurfuerst:
@hlubek full ack to what you wrote; would be really nice to do it at query-time.

Currently, I'm just having a project with 50+ editors and no dimensions; and there it is just not feasible at all to index everything 50 times. It's fine if they do not have search available in the BE for them....

What about adding these notes to the Settings.yaml file (so people know about the drawbacks if they switch this setting) and then merging it?

All the best,
Sebastian

kitsunet:
If properly documented I guess that is fine as a quick fix for now.

PR #64, opened 2014-12-22
[FEATURE] Basic support for aggregations

aertmann:

PR #91, opened 2015-09-24
[TASK] Querybuilder refactoring

daniellienert:
Refactor the queryBuilder and the result object as a basis for further extensions and improvements.
ElasticSearchQueryBuilder::fetch() now returns an array including the nodes and the other meta data results. The ElasticSearchQueryResult still provides the nodes through its ArrayAccess implementation but also returns additional data like aggregations and suggestions through specific getters.

Also added a first functional test. All tests are failing right now due to existing bugs.

Jira: IN-4

### DIFF
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
@@ -512,7 +512,7 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
 	 * @return integer
 	 */
 	public function getTotalItems() {
-		return $this->result['totalItemCount'];
+		return $this->result['total'];
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -561,64 +555,27 @@ public function fetch() {
...
 
 		if ($this->logThisQuery === TRUE) {
-			$this->logger->log('Query Log (' . $this->logMessage . ') Number of returned results: ' . count($nodes), LOG_DEBUG);
+			$this->logger->log(sprintf('Query Log (%s): %s -- execution time: %s ms -- Limit: %s -- Number of results returned: %s -- Total Results: %s',
+				$this->logMessage, json_encode($this->request), (($timeAfterwards - $timeBefore) * 1000), $this->limit, count($hits['hits']), $hits['total'])
+				, LOG_DEBUG);
 		}
 
-		$this->elasticSearchHitsIndexedByNodeFromLastRequest = $elasticSearchHitPerNode;
+		$this->result['totalItemCount'] = $hits['total'];
```
skurfuerst: any reason why we convert the result to a different format and not just say $this->result = $hits?

skurfuerst:
Hey Daniel,

See my nitpick :-) It would be great to have a small clarification whether the "array-based API" should be public or not; I suppose it should NOT be public. Correct?

All the best,
Sebastian

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -751,4 +702,53 @@ public function query(NodeInterface $contextNode) {
 	public function allowsCallOfMethod($methodName) {
 		return TRUE;
 	}
+
+	/**
+	 * @param $hits
```
aertmann: missing type..

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -751,4 +702,53 @@ public function query(NodeInterface $contextNode) {
...
+	 * @return array
+	 */
+	protected function convertHitsToNodes($hits) {
```
aertmann: missing type hinting

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -561,64 +555,27 @@ public function fetch() {
...
+				$this->logMessage, json_encode($this->request), (($timeAfterwards - $timeBefore) * 1000), $this->limit, count($hits['hits']), $hits['total'])
+				, LOG_DEBUG);
 		}
 
-		$this->elasticSearchHitsIndexedByNodeFromLastRequest = $elasticSearchHitPerNode;
+		$this->result['totalItemCount'] = $hits['total'];
 
-		if(array_key_exists("aggregations", $treatedContent)) {
-			$this->elasticSearchAggregationsFromLastRequest = $treatedContent['aggregations'];
+		if(array_key_exists('aggregations', $treatedContent)) {
```
aertmann: missing space

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
daniellienert: It should rather be $this->result = $treatedContent as we want to return the complete result including the e.g. aggregates. But that would fine.

aertmann:
The direction looks good. As we discussed yesterday I'd still like to make it easier to achieve two things.

Getting the meta result, which is currently not possible as I see it. You can only get certain things like aggregations. Having methods for specific meta results could be okay, however there should be a way to get the whole raw result and then it might not make sense having convenience methods like getAggregations but just use getResult()['aggregations'] (the treatedContent from the ES result).

Second one use case is looping over the result and getting the meta information for the hit. Currently that requires a method call to the result or query, which is cumbersome to do and impossible in Fluid. I'd suggest to make it possible to get a preprocessed array where each entry is an array consisting of a node key with the node and a meta key with the hit information. Similar to https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/64/files#diff-ccbfdb75d7efc0a0a8fe45d0ecfa5981R408

aertmann:
Addendum on last comment, discussed with Daniel and since ES only seems to have two special return keys ``aggregations`` and ``suggest`` it can be okay to have convenience methods for those, however internally I'd still suggest to not make that explicit and store the whole result and use that instead.

Issue #92, opened 2015-09-24
Travis CI doesn't test pull request


hlubek:
See the Neos or Flow development collections for a workaround to test pull requests and post merges.

johannessteu:
Gonna do this with another fix regarding functional tests

PR #96, opened 2015-09-24
[TASK] Enable elasticsearch for functional tests in travis

johannessteu:
Fixes: IN-8
Closes: #92

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,23 @@ matrix:
   include:
     - php: 5.6
     - php: 5.5
-# use modern infrastructure based on containers  
 sudo: false
 before_install:
+  - cd ..
+  - wget --no-check-certificate https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.0.zip && unzip elasticsearch-1.7.0.zip
+  - mv elasticsearch-1.7.0 elasticsearch
+  - cd elasticsearch
+  - bin/elasticsearch -d
   - cd ..
   - git clone https://github.com/neos/neos-base-distribution.git
   - cd neos-base-distribution
   - composer require flowpack/elasticsearch-contentrepositoryadaptor dev-master
 install:
   - composer install
+  - cd ..
+  - rm -rf neos-base-distribution/Packages/Application/Flowpack.ElasticSearch.ContentRepositoryAdaptor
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -3,14 +3,25 @@ matrix:
...
   - cd neos-base-distribution
   - composer require flowpack/elasticsearch-contentrepositoryadaptor dev-master
 install:
   - composer install
+  - cd ..
+  - rm -rf neos-base-distribution/Packages/Application/Flowpack.ElasticSearch.ContentRepositoryAdaptor
+  - mv Flowpack.ElasticSearch.ContentRepositoryAdaptor neos-base-distribution/Packages/Application/Flowpack.ElasticSearch.ContentRepositoryAdaptor
+  - cd neos-base-distribution
+before_script:
+  - sleep 15
```
daniellienert: Is the sleep still needed?

Tests/Functional/Eel/ElasticSearchQueryTest.php
```diff
@@ -57,26 +62,30 @@ class ElasticSearchQueryTest extends \TYPO3\Flow\Tests\FunctionalTestCase
...
 
+        $this->createNodesForNodeSearchTest();
     }
 
-
     public function tearDown()
     {
         parent::tearDown();
         $this->inject($this->contextFactory, 'contextInstances', array());
-        $this->nodeIndexCommandController->cleanupCommand();
```
daniellienert: Shouldn't we keep that cleanUp?

daniellienert:
Apart from the comments above - it is pretty cool to have functional tests for the package. This is a great start for testing and improving the package!

.travis.yml
johannessteu: I'll try some runs without the sleep. Well, the service will be started before but it might be possible that it is not up and running already.

Tests/Functional/Eel/ElasticSearchQueryTest.php
johannessteu: Since we have to use the build() command on each test there is no need to clean up the index, right?

daniellienert: I don't think so - every Build() Command will create a new index suffixed with the current timestamp and switches the alias. That doesn't hurt in the travis environment, but when you run the tests locally it does.

daniellienert: By the way - we need a check for a running elasticsearch server during the setup and skip the tests if we do not find one....

PR #97, opened 2015-09-25
[BUGFIX] If no aggregations are set getAggregations will brake

johannessteu:
If you don't request aggregations in your Query but you do access the aggregations in Fluid it results in an Exception. This change will return an empty array if there are no aggregations.

Fixes: IN-7


daniellienert:
:+1:

PR #100, opened 2015-10-06
[TASK] Add generic sort operation

johannessteu:
This changes adds a generic sort operation that can be used to create any kind of sorting that ElasticSearch supports.

### DIFF
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
@@ -171,18 +171,14 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
      */
     public function sortDesc($propertyName)
     {
-        if (!isset($this->request['sort'])) {
-            $this->request['sort'] = array();
-        }
+        $configuration = [
+            $propertyName => ['order' => 'desc']
+        ];
 
-        // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-sort.html
-        $this->request['sort'][] = array(
-            $propertyName => array('order' => 'desc')
-        );
+        $this->sort($configuration);
 
         return $this;
     }
-
 
     /**
      * Sort ascending by $propertyName
@@ -193,18 +189,32 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
      */
     public function sortAsc($propertyName)
     {
-        if (!isset($this->request['sort'])) {
-            $this->request['sort'] = array();
-        }
+        $configuration = [
+            $propertyName => ['order' => 'asc']
+        ];
 
-        // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-sort.html
-        $this->request['sort'][] = array(
-            $propertyName => array('order' => 'asc')
-        );
+        $this->sort($configuration);
 
         return $this;
     }
 
+    /**
+     * Add a $configuration sort filter to the request
+     *
+     * @param array $configuration
+     * @return ElasticSearchQueryBuilder
+     * @api
+     */
+    public function sort($configuration)
+    {
+        if (!isset($this->request['sort'])) {
+            $this->request['sort'] = array();
+        }
+
+        $this->request['sort'][] = $configuration;
+
+        return $this;
+    }
 
     /**
      * output only $limit records
@@ -451,7 +461,7 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
      * nodes = ${Search....aggregation("color", this.aggregationDefinition).execute()}
      *
      * Access all aggregation data with {nodes.aggregations} in your fluid template
-     * 
+     *
      * @param string $name
      * @param array $aggregationDefinition
      * @param null $parentPath
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryResult.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryResult.php
@@ -192,7 +192,7 @@ class ElasticSearchQueryResult implements QueryResultInterface, ProtectedContext
     public function getAggregations()
     {
         $this->initialize();
-        if (array_key_exists("aggregations", $this->result)) {
+        if (array_key_exists('aggregations', $this->result)) {
             return $this->result['aggregations'];
         }
         return array();
@@ -211,6 +211,23 @@ class ElasticSearchQueryResult implements QueryResultInterface, ProtectedContext
     public function searchHitForNode(NodeInterface $node)
     {
         return $this->elasticSearchQuery->getQueryBuilder()->getFullElasticSearchHitForNode($node);
+    }
+
+    /**
+     * Returns the array with all sort values for a given node. The values are fetched from the raw content
+     * ElasticSearch returns within the hit data
+     *
+     * @param NodeInterface $node
+     * @return array
+     */
+    public function getSortValuesForNode(NodeInterface $node)
+    {
+        $hit = $this->searchHitForNode($node);
+        if (is_array($hit) && array_key_exists('sort', $hit)) {
+            return $hit['sort'];
+        }
+
+        return array();
     }
 
     /**
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/ViewHelpers/GetHitArrayForNodeViewHelper.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/ViewHelpers/GetHitArrayForNodeViewHelper.php
@@ -1,0 +1,58 @@
+<?php
+namespace Flowpack\ElasticSearch\ContentRepositoryAdaptor\ViewHelpers;
+
+/*                                                                                                  *
+ * This script belongs to the TYPO3 Flow package "Flowpack.ElasticSearch.ContentRepositoryAdaptor". *
+ *                                                                                                  *
+ * It is free software; you can redistribute it and/or modify it under                              *
+ * the terms of the GNU Lesser General Public License, either version 3                             *
+ *  of the License, or (at your option) any later version.                                          *
+ *                                                                                                  *
+ * The TYPO3 project - inspiring people to share!                                                   *
+ *                                                                                                  */
+
+use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Eel\ElasticSearchQueryResult;
+use TYPO3\TYPO3CR\Domain\Model\NodeInterface;
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+/**
+ * View helper to get the raw "hits" array of an ElasticSearchQueryResult for a
+ * specific node.
+ *
+ * = Examples =
+ *
+ * <code title="Basic usage">
+ * {esCrAdapter:geHitArrayForNode(queryResultObject: result, node: node)}
+ * </code>
+ * <output>
+ * array
+ * </output>
+ *
+ * You can also return specific data
+ * <code title="Fetch specific data">
+ * {esCrAdapter:geHitArrayForNode(queryResultObject: result, node: node, path: 'sort')}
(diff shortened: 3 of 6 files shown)


johannessteu:
Waiting for #101 before this can get in

aertmann:
Hey Johannes, you're submitting code that doesn't comply with PSR-2 which the package was changed to. Maybe you could maybe set your IDE to follow it or use the parser mentioned in https://discuss.neos.io/t/psr-2-change-for-flow-and-neos/493

johannessteu:
Yeah, ide should already be on psr-2. I'll double check later why some changes are not made.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryResult.php
```diff
@@ -214,6 +214,23 @@ public function searchHitForNode(NodeInterface $node)
     }
 
     /**
+     * Returns the array with all sort values for a given node. The values are fetched from the raw content
+     * ElasticSearch returns within the hit data
+     *
+     * @param NodeInterface $node
+     * @return array
+     */
+    public function getSearchValuesForNode(NodeInterface $node)
```
daniellienert: The behavior of this method does not fit to its name ...
Do we really need this method at all?

daniellienert:
@johannessteu I think the way to request specific meta data of a node by viewHelpers is not very intuitive. Wouldn't it be a better way to provide an array of nodes including their metaData like sort form  `ElasticSearchQueryResult::$result` and iterate over this array?

johannessteu:
I think the main problem is the pagination. If you paginate over your ```ElasticSearchQueryResult``` you don't really have access to the result object. All iteration methods have to return an array with that node plus meta data. That would be a big breaking change. That's why there is the method getSearchValuesForNode, wich will be renamed, at all. There is no way to access the data. At least that is what @skurfuerst and i came up with. To get that data we thought about an ViewHelper since it is the only way to call a method and pass the current node to get meta data for that node.

I'll agree on the point that we could just provide one VH to return all meta data or with limit it with an argument to just one type or so. What do you think @daniellienert ?

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryResult.php
```diff
@@ -214,6 +214,23 @@ public function searchHitForNode(NodeInterface $node)
...
+     * Returns the array with all sort values for a given node. The values are fetched from the raw content
+     * ElasticSearch returns within the hit data
+     *
+     * @param NodeInterface $node
+     * @return array
+     */
+    public function getSortValuesForNode(NodeInterface $node)
+    {
+        $hit = $this->searchHitForNode($node);
+        if (is_array($hit) && array_key_exists("sort", $hit)) {
```
aertmann: should be single quotes

aertmann:
+1 regarding one VH instead of multiple specific ones

daniellienert:
@johannessteu - ok I haven't thought about pagination. But isn't it possible to iterate over something like ElasticSearchQueryResult->RawResults and use RawResults.node to display that node?

If not, I am also in favor of having only one viewHelper.

johannessteu:
I don't think so. It would be still at least a breaking change.
I'm not sure if it is possible to use the result as it is and set an iterator and use that interator.index to get data of an ElasticSearchQueryResult->RawResults. Still feels kind of nasty. I guess a single VH to render meta data for a node is way more flexible and could be used for some other use cases as well. Not just pagination. If you paginate over your result and below of above your list you like to show some meta data the VH could also be used. There might be a lot of use cases.

daniellienert:
Ok, than lets do it with the viewHelper.

johannessteu:
Ok, i'll try to get it in asap

johannessteu:
@daniellienert @aertmann any ideas what exactly that one VH should return? The problem ist that all sort data will be inside the hit array, aggregations, suggestions etc. will be outside the hit array. What about to just keep the ```GetHitArrayForNodeViewHelper``` as sort data can be fetched over that?

aertmann:
@johannessteu: I think that's what we both meant, remove ``GetSortValuesForNodeViewHelper`` and just use ``GetHitArrayForNodeViewHelper`` to get the sort data.

daniellienert:
Thats correct - this generic `GetHitDataForNodeViewHelper` could i my opinion have a parameter "path" to address and output a specific value.

aertmann:
would just return the array, then you can access ``{hit.sort}``

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/ViewHelpers/GetSortValuesForNodeViewHelper.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @param ElasticSearchQueryResult $queryResultObject
+     * @param NodeInterface $node
+     * @return array
+     */
+    public function render(ElasticSearchQueryResult $queryResultObject, NodeInterface $node)
+    {
+        $sort = $queryResultObject->getSortValuesForNode($node);
+
+        if (count($sort) == 1) {
```
aertmann: ===

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/ViewHelpers/GetSortValuesForNodeViewHelper.php
```diff
@@ -0,0 +1,50 @@
...
+use TYPO3\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+/**
+ * View helper to get the raw "hits" array of an ElasticSearchQueryResult for a
+ * specific node.
+ *
+ * = Examples =
+ *
+ * <code title="Basic usage">
+ * {esCrAdapter:getSortValuesForNode(queryResultObject: {result}, node: {node})}
```
aertmann: syntax is incorrect, should be ``queryResultObject: result, node: node)``

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/ViewHelpers/GetHitArrayForNodeViewHelper.php
```diff
@@ -0,0 +1,44 @@
...
+use TYPO3\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+/**
+ * View helper to get the raw "hits" array of an ElasticSearchQueryResult for a
+ * specific node.
+ *
+ * = Examples =
+ *
+ * <code title="Basic usage">
+ * {esCrAdapter:geHitArrayForNode(queryResultObject: {result}, node: {node})}
```
aertmann: syntax is incorrect, should be ``queryResultObject: result, node: node)``

johannessteu:
So here is the rework. I added the path thingy because you won't be really able to use {hit.sort}. You would have to use AliasView-VH or something. This might be more easy to use

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/ViewHelpers/GetHitArrayForNodeViewHelper.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param ElasticSearchQueryResult $queryResultObject
+     * @param NodeInterface $node
+     * @param string $path
+     * @return array
+     */
+    public function render(ElasticSearchQueryResult $queryResultObject, NodeInterface $node, $path = '')
+    {
+        $hitArray = $queryResultObject->searchHitForNode($node);
+
+        if (!empty($path) && array_key_exists($path, $hitArray)) {
```
daniellienert: I would suggest to use \TYPO3\Flow\Utility\Arrays::getValueByPath() to retrieve the array value.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -171,15 +171,12 @@ public function nodeType($nodeType)
      */
     public function sortDesc($propertyName)
     {
-        if (!isset($this->request['sort'])) {
-            $this->request['sort'] = array();
-        }
-
-        // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-sort.html
-        $this->request['sort'][] = array(
+        $configuration = array(
```
daniellienert: Use PHP 5.5 array structure?

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -193,19 +190,33 @@ public function sortDesc($propertyName)
      */
     public function sortAsc($propertyName)
     {
+        $configuration = array(
```
daniellienert: Use PHP 5.5 array structure?

Tests/Functional/Eel/ElasticSearchQueryTest.php
```diff
@@ -158,9 +158,21 @@ public function nodesWillBeSortedDesc()
      */
     public function nodesWillBeSortedAsc()
     {
-        $ascendingResult = $this->queryBuilder->query($this->context->getRootNode())->sortAsc("title")->execute();
+        $ascendingResult = $this->queryBuilder->query($this->context->getRootNode())->sortAsc('title')->execute();
         $node = $ascendingResult->getFirst();
-        $this->assertEquals("chicken", $node->getProperty("title"), "Asserting a asc sort order by property title");
+        $this->assertEquals('chicken', $node->getProperty('title'), "Asserting a asc sort order by property title");
```
daniellienert: There are still some doublequotes

Tests/Functional/Eel/ElasticSearchQueryTest.php
```diff
@@ -148,9 +148,9 @@ public function fieldBasedAggregations()
      */
     public function nodesWillBeSortedDesc()
     {
-        $descendingResult = $this->queryBuilder->query($this->context->getRootNode())->sortDesc("title")->execute();
+        $descendingResult = $this->queryBuilder->query($this->context->getRootNode())->sortDesc('title')->execute();
         $node = $descendingResult->getFirst();
-        $this->assertEquals("egg", $node->getProperty("title"), "Asserting a desc sort order by property title");
+        $this->assertEquals('egg', $node->getProperty('title'), "Asserting a desc sort order by property title");
```
daniellienert: There are still some doublequotes

Tests/Unit/ViewHelpers/GetHitArrayForNodeViewHelperTest.php
```diff
@@ -0,0 +1,103 @@
...
+ *                                                                                                  */
+
+use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Eel\ElasticSearchQueryResult;
+use Flowpack\ElasticSearch\ContentRepositoryAdaptor\ViewHelpers\GetHitArrayForNodeViewHelper;
+use TYPO3\TYPO3CR\Domain\Model\NodeInterface;
+
+/**
+ * Testcase for ElasticSearchQueryBuilder
+ */
+class GetHitArrayForNodeViewHelperTest extends \TYPO3\Flow\Tests\UnitTestCase
```
daniellienert: Also use the PHP 5.5 array structure in the test?

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -171,14 +171,11 @@ public function nodeType($nodeType)
...
     public function sortDesc($propertyName)
     {
-        if (!isset($this->request['sort'])) {
-            $this->request['sort'] = array();
-        }
-
-        // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-sort.html
-        $this->request['sort'][] = array(
+        $configuration = [
             $propertyName => array('order' => 'desc')
```
daniellienert: Mixed style makes it worse :)

johannessteu: Well, we should go through all of the code at all in another PR to refactor all arrays or not?

PR #107, opened 2015-11-13
[TASK] Add a runtime QueryResult Cache

dfeyer:
A query with the same payload will return the same content from ES, this change adds a runtime cache and implement the support of the ```$cacheResult``` in ```QueryInterface::execute```.

This change also enable the runtime cache by default in the ```ElasticSearchQueryBuilder```


daniellienert:
:+1: by reading

dfeyer:
:+1:

skurfuerst:
:+1: by reading

PR #110, opened 2015-11-30
TASK: Allow to set Filter execution for terms filter type

dfeyer:


aertmann:
:+1: by reading

dfeyer:
Will rebase this on top of https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/114 when it's merged

daniellienert:
:+1: by reading
It would be cool if you could extend the ElasticSearchQueryBuilderTest with this option.

dfeyer:
Not sure if this is really needed as #114 offer more flexibility without making the API more complexe.

kdambekalns:
Well, not sure either. Is what this tries to solve also possible now (with #114 merged)? Then we should just close this one…

dfeyer:
@kdambekalns I agree, we can close this one

PR #111, opened 2015-12-03
BUGFIX: Retrieval of aliases fails with newer API versions

robertlemke:
The NodeIndexer tries to retrieve the list of index aliases with a
GET request like `http://localhost:9200/*/_alias/typo3cr´. However,
newer API versions don't seem to support this URL format and fail with
a PatternSyntaxException (Dangling meta character '*').

This fix changes the pattern to `http://localhost:9200/_alias/typo3cr`
which seems to work with both, newer and older API versions.

Tested with Elasticsearch 1.5.2 and 1.7.2.


dfeyer:
:+1: by reading ...

aertmann:
:+1: by reading

PR #116, opened 2016-02-05
BUGFIX: Fulltext parts are lost on document update

### LINKED ISSUE #115: Fulltext parts are lost on document update
Whenever a document node itself is indexed (again), the document in the index is deleted, thus removing all existing `__fulltextParts` and by this removing the child nodes' content from the fulltext index.

The reason: `the str_replace()` on the node type name is plain wrong, it would have been better to use `NodeTypeMappingBuilder::convertNodeTypeNameToMappingName()` - that would have used `-` instead of `/`.

A regression introduced with #76.

kdambekalns:
Use proper way of mapping node type name to `_type` in index.

Closes #115


dfeyer:
:+1:

PR #117, opened 2016-03-09
TASK: Proper escaping for slash in full text query

dfeyer:
Without escaping ElasticSearch return an error for the current query. See https://github.com/elastic/elasticsearch/issues/2980


dfeyer:
Retarget this PR to branch 2.0

PR #120, opened 2016-03-15
TASK: Proper escaping for slash in full text query

dfeyer:
Without escaping ElasticSearch return an error for the current query.

See https://github.com/elastic/elasticsearch/issues/2980


kdambekalns:
Does it really have to be `\\/`? In the ticket you linked, this is needed to avoid the shell eating the backslash, but in the single-quoted string we build in PHP…

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -715,7 +715,7 @@ public function fulltext($searchWord)
     {
         $this->appendAtPath('query.filtered.query.bool.must', array(
             'query_string' => array(
-                'query' => $searchWord
+                'query' => str_replace('/', '\\/', $searchWord)
```
aertmann: hmm maybe it's actually needed to pass it through `json_encode` instead? this seems hacky at best

dfeyer:
I will push a new version. The first version was an ugly quick fix :(

kdambekalns:
Great, looking forward to it.

PR #123, opened 2016-03-22
TASK: Call method getRequest to make it more easy to extend

dfeyer:
This change replace $this->request by $this->getRequest() call during
request execution. This is useful when creating custom query builder
integration that can modifiy the request before execution.


kitsunet:
Looks nice but what about StyleCI and travis?

dfeyer:
@kitsunet Travis failed ... for a really strange reason need to check locally if I can reproduce

PR #124, opened 2016-04-24
BUGFIX: Index document titles by default

jrenggli:
The nodeType `TYPO3.Neos:Document` offers a property `title`. By adding
a `elasticSearchMapping` for this we ensure it is searchable and has the
same priority (`boost`) as `<h1>` titles in the content.

### DIFF
--- a/Configuration/NodeTypes.yaml
+++ b/Configuration/NodeTypes.yaml
@@ -116,6 +116,13 @@
     fulltext:
       isRoot: true
   properties:
+    'title':
+      search:
+        elasticSearchMapping:
+          type: string
+          include_in_all: true
+          boost: 20
+
     '__fulltextParts':
       search:
         elasticSearchMapping:


Configuration/NodeTypes.yaml
```diff
@@ -177,3 +177,12 @@
...
         fulltextExtractor: '${Indexing.extractHtmlTags(value)}'
+
+'TYPO3.Neos:Document':
+  properties:
+    'title':
+      search:
+        elasticSearchMapping:
+          type: string
+          include_in_all: true
+          boost: 20
```
aertmann: this will override the existing configuration for `TYPO3.Neos:Document`, it needs to be moved up to the first definition instead

jrenggli: Hi Aske.
Thanks for the hint. I moved it up now.

kdambekalns:
Hah, noticed this the other day as well… but, two things:

- this should go into the 2.0 branch
- why not just add the title to the fulltext?
```
    search:
        fulltextExtractor: ${Indexing.extractInto('h1', value)}
```

dimaip:
Closing in favor of #152. Thanks for the PR, @jrenggli!

PR #127, opened 2016-06-20
TASK: Free memory during indexing after each dimension combination

nezaniel:
This prevents memory being flooded with all node / dimension combinations data during indexing by clearing caches when a dimension combination was completely indexed


kdambekalns:
Shouldn't this go into the 2.0 branch? Or does it affect only master?

kdambekalns:
Created #135 against the 2.0 branch.

PR #133, opened 2016-06-30
TASK: Add LICENSE, adjust composer manifest

kdambekalns:


kdambekalns:
The dependency on typo3/eel is removed, since the package depends on flowpack/elasticsearch, which depends on typo3/flow, which depends on typo3/eel. :)

PR #135, opened 2016-07-06
TASK: Free memory during indexing after each dimension combination

kdambekalns:
This prevents memory being flooded with all node / dimension combinations data during indexing by clearing caches when a dimension combination was completely indexed


kdambekalns:
@nezaniel

kdambekalns:
Currently running `nodeindex:build` here, 3373 nodes and two dimensions with ~10 values each. Stays at ~2.3 GB of RAM with this change.

aertmann:
👍 by reading

PR #136, opened 2016-07-06
TASK: Implement BulkNodeIndexerInterface in NodeIndexer

kdambekalns:
This change implements the newly introduced `BulkNodeIndexerInterface` for
the TYPO3.TYPO3CR.Search package. When enabled this skips the check for
duplicate nodes with a different node type. This check is only useful when
changing the node type (live indexing). During the initial indexing this can
speed up the process quite a lot for big indexes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_install:
   - git clone https://github.com/neos/neos-base-distribution.git
   - cd neos-base-distribution
   - composer require flowpack/elasticsearch-contentrepositoryadaptor dev-master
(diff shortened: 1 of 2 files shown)


kdambekalns:
Hey, @dfeyer - the tests pass now! :)

kdambekalns:
Currently indexing 3373 nodes and that takes ~10 minutes _without_ this change. I'll apply it and report back on the (hopefully) improved speed later.

kdambekalns:
So this change brings down the time to ~4 minutes for each set of ~3370 nodes in my setup.

Issue #139, opened 2016-07-21
ElasticSearchQueryBuilder::getTotalItems() always return NULL


bwaidelich:
`ElasticSearchQueryBuilder::getTotalItems()` checks for a key `total` in the result but that never exist. Instead the total hits are returned in `$this->results['hits']['total']` instead.

So shouldn't:
```php
    public function getTotalItems()
    {
        if (array_key_exists('total', $this->result)) {
            return (int)$this->result['total'];
        }
    }
```

be changed to sth like:

```php
    public function getTotalItems()
    {
        if (isset($this->result['hits']['total'])) {
            return (int)$this->result['hits']['total'];
        }
        return 0;
    }
```
?

bwaidelich:
If someone ( @skurfuerst @dfeyer ?) can confirm this, I'll happily prepare a PR

dfeyer:
It's really weird ... I will check you fix when I'm back from holiday, if I forgot, ping me on slack.

dfeyer:
@bwaidelich Can you rebase your PR and I think we can merge it

bwaidelich:
@dfeyer done, see #140 thx for checking

PR #140, opened 2016-07-22
BUGFIX: Correctly determine number of total search results

bwaidelich:
This fixes `ElasticSearchQueryBuilder::getTotalItems()` that was checking
a non-existing result index `total` to retrieve the number of search results.

Fixes: #139


dfeyer:
Thanks, its definitively a bugfix, so maybe we can target branch 2.0 ?

bwaidelich:
> so maybe we can target branch 2.0 ?

Sure, feel free ;)
I'll take care

bwaidelich:
See #144

PR #144, opened 2016-08-10
BUGFIX: Correctly determine number of total search results

bwaidelich:
This fixes ElasticSearchQueryBuilder::getTotalItems() that was checking
a non-existing result index total to retrieve the number of search results.

Fixes: #139


bwaidelich:
See #140 for the original PR against master

PR #145, opened 2016-09-05
TASK: make uriPathSegment not_analyzed to allow term matching

dimaip:


dfeyer:
I think this one can be target to branch 2.0 ?

PR #118, opened 2016-03-15
TASK: Add support for ElasticSearch 2.*

dfeyer:
This change add support for recent version of ElasticSearch. This change is not backward compatible and break the support for earlier version.

PR #149, opened 2016-10-04
BUGFIX: Include the context document in the query result

dfeyer:
The ElasticSearchQueryBuilder::query searches only below the given context
document.

One example when this is wrong is when you search from the site node, which
you use as the homepage - in this case the actual "homepage" would not be
searched.

### DIFF
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
@@ -777,7 +777,16 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
         // on indexing, the __parentPath is tokenized to contain ALL parent path parts,
         // e.g. /foo, /foo/bar/, /foo/bar/baz; to speed up matching.. That's why we use a simple "term" filter here.
         // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-term-filter.html
-        $this->queryFilter('term', ['__parentPath' => $contextNode->getPath()]);
+        $this->queryFilter('bool', [
+            'should' => [
+                [
+                    'term' => ['__parentPath' => $contextNode->getPath()]
+                ],
+                [
+                    'term' => ['__path' => $contextNode->getPath()]
+                ]
+            ]
+        ]);
 
         //
         // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-terms-filter.html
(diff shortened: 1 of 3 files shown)


dfeyer:
Don't merge this one in the 2.0 branch, the query syntax is only supported by 2.0+ :(

dfeyer:
I force push a fix to support ES 1.7 -> 2.4, basically minimum_should_match was not needed because the default value is 1

hoodini:
made composer update with neos and used this changes directly to Flowpack.ElasticSearch.ContentRepositor, it works perfect!

dimaip:
@dfeyer I can verify that this PR fixes the bug. But why are the tests failing?

dfeyer:
@dimaip Because the query structure is not the same I push a fix, and we can merge this one as soon as travis is green and release a minor version

kdambekalns:
Took the freedom to edit title and description a  bit. And I dare to oppose - you do specify the _context node_, below which you want to search. That's how it worked so far.

Now this would change the behaviour, potentially doing awkward things to search results. No?

As far as I understand, the "site node is the homepage" is kind of an edge case here…

dfeyer:
Right ...

dfeyer:
Retargeted to master, test will fail before merging #118

dfeyer:
@kdambekalns If we document this change in the release note of the next major (3.0) is it OK for you ? We can even document how to get back to the old behaviour by adding a filter to ignore the current node ?

kdambekalns:
Sure, no problem then.

kdambekalns:
So, happily updated to 3.0 and was disappointed, since this still didn't work. Checked the code, and the inclusion of the context node into the search is not in the 3.0 codebase. It probably got lost during the driver rewrite?!

Any insights, @dfeyer or @daniellienert maybe?

daniellienert:
@kdambekalns: Sorry, that one slipped through. The tests in current master asserts 3 nodes whereas there should be 4.

kdambekalns:
Now, what do we do? This change was pushed back to 3.0, since it was considered breaking. Now 3.0 and 4.0 are released, does this mean it has to go into 5.0?

kitsunet:
So, uhm, How is site node being the homepage an edge case? It's the default we deliver with neos, so it's certainly not an edge case.

I consider this a serious bug. How about introducing a second argumetn to `query` enabling this and immediately deprecating this flag to be defaulted to true and then removed in next major.

kdambekalns:
Edge case: most sites I know make the site node a shortcut, to ease navigation building. But whatever, that's no longer important. ;)

A flag? Hm, not very elegant.

kdambekalns:
@dfeyer suggested

> We can even document how to get back to the old behaviour by adding a filter to ignore the current node ?

How would that filter look like?

PR #154, opened 2016-10-20
BUGFIX: Correctly handle removed/hidden nodes in fulltext index

kdambekalns:
These changes make sure the fulltext index is correctly updated when nodes are hidden or removed.


dfeyer:
Nice one, tested on two production project, look like it does not break any thing.

PR #155, opened 2016-10-24
BUGFIX: Unneeded dependency on TYPO3.Neos package

radmiraal:
The contentDimensionPresetSource was typehinted to the Neos
ContentDimensionPresetSourceInterface from the TYPO3.Neos
package. This interface is based on the corresponding
ContentDimensionPresetSourceInterface interface in the TYPO3.TYPO3CR
package.

As the command controller does not make use of the additional
method on the interface in the Neos package the use of it is
useless.

This change changes the type hint to the TYPO3.TYPO3CR package
so the Elastic content repository adaptor can be installed without
the TYPO3.Neos package.

### DIFF
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Command/NodeIndexCommandController.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Command/NodeIndexCommandController.php
@@ -61,7 +61,7 @@ class NodeIndexCommandController extends CommandController
 
     /**
      * @Flow\Inject
-     * @var \TYPO3\Neos\Domain\Service\ContentDimensionPresetSourceInterface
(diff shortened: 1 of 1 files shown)


aertmann:
hey this is also in 1.2 and 2.0, shouldn't it be fixed there instead of master and upmerged?

radmiraal:
@aertmann tbh I didn't even check, expected it to be a master only issue as I just switched to master while running into the issue... I can just change the target branch of the PR right?

aertmann:
@radmiraal yep, but you might need to rebase the branch after changing it

radmiraal:
k, based it on 1.2 now but that would require at least someone running the package as I don't have a 1.2 project available with the 1.2 version of the package and don't have the 'luxury of time' to set it up

dfeyer:
Test are failing because no test for branch 1.2 ... and even not sure if we can consider this branch as maintained ... let's discuss that later ;)

PR #162, opened 2016-12-19
TASK: Apply code migrations

gerhard-boden:


Classes/ViewHelpers/GetHitArrayForNodeViewHelper.php
```diff
@@ -39,6 +39,13 @@
  */
 class GetHitArrayForNodeViewHelper extends AbstractViewHelper
 {
+    
+	/**
+	 * NOTE: This property has been introduced via code migration to ensure backwards-compatibility.
+	 * @see AbstractViewHelper::isOutputEscapingEnabled()
+	 * @var boolean
+	 */
+	protected $escapeOutput = FALSE;
```
kdambekalns: I think this is not needed and has been added by an old migration…

gerhard-boden:
Thanks @kdambekalns for taking care of the change, would not have gotten to take care of it until now

baschny:
My setup was:
```
        "typo3/neos": "~2.2.0",
        "flowpack/elasticsearch-contentrepositoryadaptor": "dev-master",
        "flowpack/elasticsearch": "dev-master",
```
This was working until this PR was merged. I require ES 2.x support (which was never released from dev-master) with Neos 2.x. 

Now you rised the minimum requirements to Flow 4, but there is no release that I can use in my scenario anymore.

My suggestion would be to make a release a  2.1.0 of the state before this PR here was merged, so that we can require this in our Neos 2.3 projects (with ES 2.x).

@kdambekalns ?

baschny:
My workaround for now:

```
    "flowpack/elasticsearch-contentrepositoryadaptor": "2.0.x-dev#a5d04f6"
```

The "2.0.x-dev" seems to be required (even thou that commit is not part of that branch), so that composer checks the correct constraints - and then check out the specific commit which is the commit that I was before (with dev-master).

daniellienert:
@baschny: I am really sorry for the inconvenience. The problem is, that with the namespace changes, we would need a matrix of versions, supporting Flow 3.3 / Flow 4.0 on the one hand and ES 1.7 / 2.x and later 5.x on the other. 
That would be doable but a hell to maintain. Our solution now is to write Driver classes, that abstract the ES Version. 
We hope to come up with that solution soon. In the meanwhile your workaround would be my approach to - and is even more stable than dev-master :)

PR #169, opened 2017-01-17
BUGFIX: Unneeded dependency on TYPO3.Neos package

radmiraal:
The contentDimensionPresetSource was typehinted to the Neos
ContentDimensionPresetSourceInterface from the TYPO3.Neos
package. This interface is based on the corresponding
ContentDimensionPresetSourceInterface interface in the TYPO3.TYPO3CR
package.

As the command controller does not make use of the additional
method on the interface in the Neos package the use of it is
useless.

This change changes the type hint to the TYPO3.TYPO3CR package
so the Elastic content rpeository adaptor can be installed without
the TYPO3.Neos package.

### DIFF
--- a/Classes/Command/NodeIndexCommandController.php
+++ b/Classes/Command/NodeIndexCommandController.php
@@ -52,7 +52,7 @@ class NodeIndexCommandController extends CommandController
 
     /**
      * @Flow\Inject
-     * @var \Neos\Neos\Domain\Service\ContentDimensionPresetSourceInterface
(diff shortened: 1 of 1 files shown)


radmiraal:
This change was merged in the 1.2 branch and the problem is non-existing in the 2.0 branch. But for some reason it did not end up in master.

dfeyer:
Yes master is currently out of sync due to big merge conflict 2.0 -> master :(

PR #172, opened 2017-01-25
TASK: Update documentation about custom index name

chrschw:


dfeyer:
Hi @chrschw thanks for your PR. It's not exactly MultiDomain support, because currently Multi domain support in Neos is something different, that we don't do right in Elastic right now, check #177

PR #179, opened 2017-02-03
BUGFIX: Change wrong variable name

Nikdro:


dfeyer:
Thanks for this one. Was introduced when I refactor the method name

PR #180, opened 2017-02-04
TASK: Refactor the object factory

daniellienert:
Separate driver and query factory


daniellienert:
@dfeyer Just a small refactoring.

PR #183, opened 2017-02-09
BUGFIX: Make indexing work on hidden content

radmiraal:
The node indexing uses a content context which does not
return hidden content. This means hidden nodes are
not imported while the _hidden property is mapped and
excluded from queries by default.

### DIFF
--- a/Classes/Command/NodeIndexCommandController.php
+++ b/Classes/Command/NodeIndexCommandController.php
@@ -249,7 +249,11 @@ class NodeIndexCommandController extends CommandController
     protected function indexWorkspaceWithDimensions($workspaceName, array $dimensions = [])
     {
         $indexedNodes = 0;
-        $context = $this->contextFactory->create(['workspaceName' => $workspaceName, 'dimensions' => $dimensions]);
+        $context = $this->contextFactory->create([
+            'workspaceName' => $workspaceName,
+            'dimensions' => $dimensions,
+            'invisibleContentShown' => true
+        ]);
         $rootNode = $context->getRootNode();
 
         $indexedNodes += $this->traverseNodes($rootNode);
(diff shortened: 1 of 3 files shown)


dfeyer:
The discussion in Slack https://neos-project.slack.com/archives/guild-search/p1486821186000518

Now hidden and deleted nodes are not indexed at all and destroyed by the NodeIndexer https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/2.0/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Indexer/NodeIndexer.php#L355

We must remove this configuration, non sense ... and we need also to update the query if hidden node are not in the index (edited)

Currently this behaviour is hardcoded, but with the upcoming Elastic Driver support, the groovy script can use custom implementation, so you if this behaviour is wrong for you, you can change it

kdambekalns:
@dfeyer as discussed, I know of projects where `_hidden` is actually used in queries. I still think those nodes end up in the index through realtime indexing, since that just passes published nodes, not fetching them through a Context instance.

So, IMHO this change totally makes sense.

How this is changed by the driver concept, I have no clue, though. Can you elaborate on that, maybe?

kdambekalns:
And getting back to line 355 in the indexer: Hidden nodes must not be in the aggregated fulltext content for a _fulltext root node_. That's all that this line does, it does not "destroy nodes".

Granted, being able to find nodes based on content of child nodes depending on the hidden state of those very child nodes would be awesome. But that line fixed a very nasty bug.

So, I think you mix things up here?!

dfeyer:
> So, I think you mix things up here?!

Yes yes ;)

kdambekalns:
I guess rebasing this on 3.0 is sufficient…

kdambekalns:
Does anyone have a clue on the state/usefulness of this one?

PR #184, opened 2017-02-19
!!! TASK: Adjust the result of getSuggestions

daniellienert:
Before the result of ``getSuggestions`` returned the suggestion of a search term if the term only consisted of one word. When the search term consisted of two words it returned the complete array. This is an absolutely unexpected behavior. It was also only possible to access the first defined suggest term.

This change now returns the complete suggestion array, so it behaves like the comparable aggregation getter.


daniellienert:
@kdambekalns: Nope, this feature is not used in the SearchPlugin. The suggestions are used for the live suggestions while typing, but this is build separately through a raw query in an ajax controller.

PR #185, opened 2017-02-19
FEATURE: Make amount of buckets configurable via fusion

daniellienert:
By default, elasticsearch returns 10 buckets for a term aggregation.
This change makes the size value configurable via fusion.

Also improves some header comments.
This is a rebuild-PR against new 3.0.

### DIFF
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Driver/QueryInterface.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Driver/QueryInterface.php
@@ -91,14 +91,14 @@ interface QueryInterface
     /**
      * This method is used to create any kind of aggregation.
      *
-     * @param string $name
+     * @param string $name The name to identify the resulting aggregation
      * @param array $aggregationDefinition
-     * @param string $parentPath
+     * @param string $parentPath ParentPath to define the parent of a sub aggregation
      * @return void
      * @api
      * @throws Exception\QueryBuildingException
      */
-    public function aggregation($name, array $aggregationDefinition, $parentPath = null);
+    public function aggregation($name, array $aggregationDefinition, $parentPath = '');
 
     /**
      * This method is used to create any kind of suggestion.
--- a/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
@@ -368,17 +368,19 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
      *
      * Access all aggregation data with {nodes.aggregations} in your fluid template
      *
-     * @param $name
-     * @param $field
-     * @param string $type
-     * @param null $parentPath
+     * @param string $name The name to identify the resulting aggregation
+     * @param string $field The field to aggregate by
+     * @param string $type Aggregation type
+     * @param string $parentPath
+     * @param int $size The amount of buckets to return
      * @return $this
      */
-    public function fieldBasedAggregation($name, $field, $type = "terms", $parentPath = null)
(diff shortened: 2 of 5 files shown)


Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Driver/QueryInterface.php
```diff
@@ -91,14 +91,14 @@ public function highlight($fragmentSize, $fragmentCount = null);
...
+     * @param string $name The name to identify the resulting aggregation
      * @param array $aggregationDefinition
-     * @param string $parentPath
+     * @param string $parentPath ParentPath to define the parent of a sub aggregation
      * @return void
      * @api
      * @throws Exception\QueryBuildingException
      */
-    public function aggregation($name, array $aggregationDefinition, $parentPath = null);
+    public function aggregation($name, array $aggregationDefinition, $parentPath = '');
```
kdambekalns: Changing the interface makes this breaking (I know you know)–but: why that change?

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -368,17 +368,19 @@ public function queryFilterMultiple($data, $clauseType = 'must')
...
-     * @param null $parentPath
+     * @param string $name The name to identify the resulting aggregation
+     * @param string $field The field to aggregate by
+     * @param string $type Aggregation type
+     * @param string $parentPath
+     * @param int $size The amount of buckets to return
      * @return $this
      */
-    public function fieldBasedAggregation($name, $field, $type = "terms", $parentPath = null)
+    public function fieldBasedAggregation($name, $field, $type = 'terms', $parentPath = '', $size = 10)
```
kdambekalns: Same here–why change the default value?

daniellienert: Lets start here. Adding the $size as last argument means that I have to give the parentPath in eel. (do we have `null` in eel?). 
If we have that - As the path is a string, a default value of null is completely unintuitive for integrators and  a '' given in eel would not work.

Thats why I changed the API here. I am open to your suggestions.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Driver/QueryInterface.php
daniellienert: See the actual background below. I change the (IMHO bad default value) consistently through the methods.

daniellienert:
@kdambekalns: I don't want to introduce breaking changes if not needed. I thought due to the `null` value it would be necessary. And in addition `''` as default fits better.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/Eel/ElasticSearchQueryBuilder.php
aertmann: null is a concept in Eel, `${null}` returns `NULL`

aertmann: question is if it should accept both '' and `null` as empty values tho

daniellienert: Hey @aertmann, thanks. That would be a okish 3rd option. Is still think give a default string with null is not intuitive. So I would leave it with `''` and make the algorithm accept both.

aertmann: right, I prefer empty strings over null values for string properties too but instead of making this breaking, just accept null as empty as well.. simply cast the value to a string value before checking if it's `if ((string)$parentPath !== '') {`

PR #189, opened 2017-02-20
TASK: Switch to PSR-4

daniellienert:
Rename classes and update composer.json


daniellienert:
Travis fail can be ignored here.

PR #199, opened 2017-03-17
BUGFIX: Include the context document in the query result

kdambekalns:
The ElasticSearchQueryBuilder::query searches only below the given context
document.

One example when this is wrong is when you search from the site node, which
you use as the homepage - in this case the actual "homepage" would not be
searched.


kdambekalns:
A fresh try of actually having #149 in the codebase ;)

The question remaining: do we consider this breaking? And is adding documentation enough to remedy this, as suggested by @dfeyer in #149:

> We can even document how to get back to the old behaviour by adding a filter to ignore the current node ?

daniellienert:
@kdambekalns: I would consider it as a bug fix. Sure it is behavior changing, but which bug fix is not. I think the question is, which behavior would you expect?

Karsten, thanks for taking care of the new bugfix!! You need to take over the test changes too.

kdambekalns:
@daniellienert that should work  now…

kdambekalns:
Hi @daniellienert - please review your approval… :)

PR #201, opened 2017-04-13
BUGFIX: Typo in Readme

kuborgh-mspindelhirn:


aertmann:
would be good to fix in stable branches too

PR #204, opened 2017-04-25
BUGFIX: Index by alias, treated content

sruegg:
If you search for index that not exists, it triggers error (and should not as the code handle the creation after that).

And sometimes (perhaps due to utilisation of Elastic Cloud), treated response is null/not array so an error is triggered. Simply ensure here a empty array is returned if treated content is not an array.


daniellienert:
Hey @sruegg, thanks for that improvement. As it is a bugfix, would you pleas target the PR to the branch 3.0.

daniellienert:
Ah, you also made a PR onto 3.0. So I close this one in favor of https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/203.

PR #206, opened 2017-04-27
BUGFIX: do not iterate over all dimensions to traverse the nodetree only once

sbruggmann:
In our opinion the we don't need the allowed dimension combinations upfront ([here](https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/3.0/Classes/Command/NodeIndexCommandController.php#L169) and [here](https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/3.0/Classes/Service/IndexWorkspaceTrait.php#L49)), because the dimensions get loaded anyway in the NodeIndexer ([here](https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/3.0/Classes/Indexer/NodeIndexer.php#L240)).

Or do we overlook something and it's not enough to deliver a proper default dimension to create the root node?

### DIFF
--- a/Classes/Command/NodeIndexCommandController.php
+++ b/Classes/Command/NodeIndexCommandController.php
@@ -141,6 +141,8 @@ class NodeIndexCommandController extends CommandController
             $workspace = 'live';
         }
 
+        $defaultDimensionCombination = $this->getDefaultDimensionCombination();
+
         $indexNode = function ($identifier, Workspace $workspace, array $dimensions) {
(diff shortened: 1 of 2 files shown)


sbruggmann:
I close this PR because with it, the indexer would ignore nodes which were created outside of the default dimension.

Currently I'm working on a new PR which allows Indexer Dimension combination presets. Then our needed behavior is configurable.

PR #212, opened 2017-05-15
TASK: Check type of __fulltextParts to ensure it is always a hash map

Torsten85:
For some reason `__fulltextParts` is sometimes an empty array which prevents the groovy script from "putting" into it. In this case, this pull requests reinitializes the `_fulltextParts`as a new hashmap.

### DIFF
--- a/Classes/Driver/Version1/IndexerDriver.php
+++ b/Classes/Driver/Version1/IndexerDriver.php
@@ -114,7 +114,7 @@ class IndexerDriver extends AbstractIndexerDriver implements IndexerDriverInterf
             [
                 // first, update the __fulltextParts, then re-generate the __fulltext from all __fulltextParts
                 'script' => '
-                    if (!(ctx._source.containsKey("__fulltextParts") && ctx._source.__fulltextParts instanceof LinkedHashMap)) {
+                    if (!(ctx._source.containsKey("__fulltextParts") && ctx._source.__fulltextParts instanceof Map)) {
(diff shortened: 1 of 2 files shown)


Classes/Driver/Version1/IndexerDriver.php
```diff
@@ -114,7 +114,7 @@ public function fulltext(NodeInterface $node, array $fulltextIndexOfNode, $targe
             [
                 // first, update the __fulltextParts, then re-generate the __fulltext from all __fulltextParts
                 'script' => '
-                    if (!ctx._source.containsKey("__fulltextParts")) {
+                    if (!(ctx._source.containsKey("__fulltextParts") && ctx._source.__fulltextParts instanceof LinkedHashMap)) {
```
skurfuerst: can we use instanceof "Map" here? Would be a little more stable and in line with Java APIs.

Classes/Driver/Version2/IndexerDriver.php
```diff
@@ -124,7 +124,7 @@ public function fulltext(NodeInterface $node, array $fulltextIndexOfNode, $targe
                     'inline' => '
                         ctx._source.__fulltext = new HashMap();
                         
-                        if (!ctx._source.containsKey("__fulltextParts")) {
+                        if (!(ctx._source.containsKey("__fulltextParts") && ctx._source.__fulltextParts instanceof HashMap)) {
```
skurfuerst: can we use instanceof "Map" here? Would be a little more stable and in line with Java APIs.

PR #216, opened 2017-06-06
!!! FEATURE: Index each dimension combination into separate index

chrschw:
I changed the indexing to separate each language into another Index. 

Reason: Suggest controller results are not separated by language


kdambekalns:
Thanks, but I think this should not go into a released branch…

Also, I am not sure on the implications this has on other use cases and parts of the system. Can you elaborate a bit more on what this does and how it affects indexing and searching?

kitsunet:
Yes, the described case can be fixed by fixing the search query done in the suggest by filtering for respective dimension(s).

kdambekalns:
@kitsunet Not sure about that, TBH. Can suggestions be limited by a query like aggregations? They depend on a context, and currently we do not put the dimensions (combination hash) into that context during indexing. So… I think that is indeed not possible right now!?

chrschw:
I tried several ways to filter (comletion, analyzer), but was not successfull. Maybe you find another way?

As far as i can remember at the Neos Conference Daniel Lienert (or Sebastian Helzle) said that the elasticsearch suggest query is not able to separate different languages in one index at all. So you have to find another solution.

daniellienert:
@kitsunet, we tried that in https://github.com/Flowpack/Flowpack.SearchPlugin/pull/26/files#diff-5306144282c6c71198c056b48780e288R104 - the suggestions can be limited to the DimensionCombination while the autocompletion can't.
Apart from that, separate indexes by dimensions would make it possible to configure language specific mappings with suitable analyzers. So it would be great to have that feature.

As Elasticsearch is able to search in several indexes at the same time, I guess it would be possible to implement that feature kind of backward compatible.

Classes/Flowpack/ElasticSearch/ContentRepositoryAdaptor/ElasticSearchClient.php
```diff
@@ -34,12 +35,26 @@ class ElasticSearchClient extends \Flowpack\ElasticSearch\Domain\Model\Client
     protected $indexName;
 
     /**
+     * The language suffix
+     *
+     * @var string
+     */
+    protected $languageDimension;
```
daniellienert: Dimensions are not only about languages, although they mostly used for that purpose. We need a solution that does not equate dimension-presets with a language.

chrschw: ok i will do a rewrite as soon as possible

dfeyer:
@chrschw First a big thanks for the work, it's really a good move

👍  for @daniellienert comments about the possibility to configure more in detail the language support of ES and I can confirm the autocompletion can not be "filtred"

I'm currently working a project where I need this. So i will test it and try to improve it to make it more generic

kdambekalns:
Indexing into separate indexes per node type might be something to go for even, considering that types will be removed from Elasticsearch eventually…

dfeyer:
> Indexing into separate indexes per node type might be something to go for even, considering that types will be removed from Elasticsearch eventually…

I see the number of indexes growing quiet fast ;) but that look like a good idea, but need more modification because the query need to be spread in multiple indexes, so first, let's try to do one index per dimension combination, and next iteration will be to support one index per node type and dimensions combinations :)

dfeyer:
Just a side node about the one index per node type, it match the move of ES6 to kill types, but it can have a big impact in term of shards, basically if you have 5 shards per index ... 200 node types, will create 1'000 shards ... and querying across shards can slow down the response time. Need to be carefully testes, and maybe we need to support both (1 index with everything, or the split) ... but not sure currently and it will make the maintenance of this package harder :(

kitsunet:
Sounds like that part needs some more conception, so lets get this split by dimension in for now and then think about a good solution for the types.

dfeyer:
I need to finish a big data import for the current project, and begin of next week I start working/testings/reviewing this PR

@chrschw or if you have time to work on it, feel free, I don't want to take over this PR ;)

chrschw:
i am currently in hollydays and will be back at 19.07. in office

chrschw:
had hard weeks before hollydays so i relly need a little timeout

chrschw:
if you are in a hurry you can take over. no problem for me

dfeyer:
@kdambekalns You add Wrong Branch, because we don't introduce new feature to 2.0 branch anymore, right ? So this one need to go the 3.0

kdambekalns:
@dfeyer I only added the "wrong branch" label, this PR is not even mine ;)

dfeyer:
So the current PR is broken for live indexing, as we iterate over all dimensions combinations in the NodeIndexer and we use a bulk request here.

dfeyer:
I create a PR on master here https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/227, should be easy to backport to 3.0 ... not sure about 2.0 branch currently

dfeyer:
> As Elasticsearch is able to search in several indexes at the same time, I guess it would be possible to implement that feature kind of backward compatible.

The current code base is really hard to make the switch between indexes ... we need a hug refactor to enable this feature without breaking everything and make the code too messy

So let's start the discussion from the beginning ;) why did we need that ?

- Suggester ? All suggester support Context (see https://www.elastic.co/guide/en/elasticsearch/reference/2.4/suggester-context.html) so I think the dimensionsHash can be used as a context, and all suggester, including Completion will work

Did you see any good reason for a big refactoring of the package ? (please say no, and use ES feature)

Next we can plan some big refactoring to support ES6 if they break the support for Types ;)

dfeyer:
I close this one in favor of #227

PR #175, opened 2017-01-27
TASK: updated travis.yml | removed any "typo3" | added all dependencies

ComiR:

PR #217, opened 2017-06-21
TASK: Update travis.yml, remove "typo3", declare dependencies

kdambekalns:


kdambekalns:
This is basically #175 cherry-picked to the current master… so, thanks to @ComiR for most of the changes!

Configuration/Settings.yaml
```diff
@@ -106,9 +106,18 @@ Flowpack:
...
+      # If you only index the live workspace, Search will not work for your editors in the user workspaces.
+      # Furthermore, if you heavily rely on Search for collecting content, this might be strange for editors to
+      # work with -- as unpublished changes are not indexed right away.
       indexAllWorkspaces: true
+
       elasticSearch:
-        indexName: typo3cr
+
+        # API. name of the Elasticsearch index to use. Will create many indices prefixed by this indexName.
+        indexName: neoscr
```
daniellienert: This one is breaking when you make your own mapping adjustments, like we did here: https://github.com/Flowpack/Flowpack.SearchPlugin/pull/26/files#diff-d06e40a2a4877cacd0ccde3ab14e8588R23
So we need some migration instructions for that.

kdambekalns: Does the (new) PR description work for you? I'd also make sure a release hints at the needed changes. Or should we stay with the "old name" to be on the safe side?

And… didn't we change that name elsewhere already anyways? Hm…

aertmann: Good catch @daniellienert. These changes doesn't really justify a breaking change IMO, so maybe keep it for now and change next time a major version is needed? e.g. open a pull request this specifically for it and label it "Next major"

kdambekalns:
Ok, I'll adjust that.

Configuration/Settings.yaml
dfeyer: Basically this PR is for master, and no planned release from master currently, we have branch 1.2 (stale), 2.0 and 3.0 (active) for stable release, so no object to merge this in master

aertmann: Good point. Does the ES 5 support need a new major @kdambekalns?

kdambekalns: No it does not IMHO, and in fact I'd love to see that merged into 3.0 as indicated on #220.

kdambekalns: But, @dfeyer what about the 4.0.x releases that have been created off of the master branch!?

aertmann: oh right, forgot there's a 4.x version

PR #220, opened 2017-06-27
FEATURE: Support for Elasticsearch v5

kdambekalns:
These changes enable support for Elasticsearch v5 via a new driver and changes
in the internal handling of building type mappings.

### DIFF
--- a/Classes/Driver/Version1/DocumentDriver.php
+++ b/Classes/Driver/Version1/DocumentDriver.php
@@ -14,7 +14,7 @@ namespace Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\Version1;
 use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\AbstractDriver;
 use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\DocumentDriverInterface;
 use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\DriverInterface;
-use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Mapping\NodeTypeMappingBuilder;
+use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\NodeTypeMappingBuilderInterface;
 use Flowpack\ElasticSearch\Domain\Model\Index;
 use TYPO3\Flow\Annotations as Flow;
 use TYPO3\TYPO3CR\Domain\Model\NodeInterface;
@@ -28,14 +28,21 @@ use TYPO3\TYPO3CR\Domain\Model\NodeType;
 class DocumentDriver extends AbstractDriver implements DocumentDriverInterface
 {
     /**
+     * @Flow\Inject
+     * @var NodeTypeMappingBuilderInterface
+     */
+    protected $nodeTypeMappingBuilder;
+
+    /**
      * {@inheritdoc}
      */
     public function delete(NodeInterface $node, $identifier)
     {
+        $nodeTypeMappingBuilder = $this->nodeTypeMappingBuilder;
         return [
             [
                 'delete' => [
-                    '_type' => NodeTypeMappingBuilder::convertNodeTypeNameToMappingName($node->getNodeType()),
+                    '_type' => $nodeTypeMappingBuilder::convertNodeTypeNameToMappingName($node->getNodeType()),
                     '_id' => $identifier
                 ]
             ]
@@ -47,6 +54,8 @@ class DocumentDriver extends AbstractDriver implements DocumentDriverInterface
      */
     public function deleteDuplicateDocumentNotMatchingType(Index $index, $documentIdentifier, NodeType $nodeType)
     {
+        $nodeTypeMappingBuilder = $this->nodeTypeMappingBuilder;
+
         $index->request('DELETE', '/_query', [], json_encode([
             'query' => [
                 'bool' => [
@@ -57,7 +66,7 @@ class DocumentDriver extends AbstractDriver implements DocumentDriverInterface
                     ],
                     'must_not' => [
                         'term' => [
-                            '_type' => NodeTypeMappingBuilder::convertNodeTypeNameToMappingName($nodeType->getName())
+                            '_type' => $nodeTypeMappingBuilder::convertNodeTypeNameToMappingName($nodeType->getName())
                         ]
                     ],
                 ]
--- a/Classes/Driver/Version5/Mapping/NodeTypeMappingBuilder.php
+++ b/Classes/Driver/Version5/Mapping/NodeTypeMappingBuilder.php
@@ -1,0 +1,208 @@
+<?php
+namespace Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\Version5\Mapping;
+
+/*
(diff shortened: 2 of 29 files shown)


kdambekalns:
This is basically two parts - the first commits are independent of v5 as such and could be verified and even merged independently. I'll gladly create a new PR for that, if desired. We could even turn them around, so that the use of indexed scripts comes first. Note: those _do_ need proper review, even though they _should_ not change behaviour. Or rather, _because_ of that.

The second part is support vor Elasticsearch v5. This has two obvious problems from my POV:

- HashMap.each() in fulltext indexing causes compilation errors
- a field type conflict with `__dimensionCombinations.language` occurs

For both cases hints are appreciated!

Classes/Driver/Version5/Mapping/NodeTypeMappingBuilder.php
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Adjust the mapping for string to text or keyword as needed.
+     *
+     * This is used to ease moving from ES 1.x and 2.x to 5.x by migrating the
+     * mapping like this:
+     *
+     * | 2.x                                       | 5.x                              |
+     * |-------------------------------------------|----------------------------------|
+     * | "type": "string", "index": "no"           | "type": "text", "index": false   |
+     * | "type": "string"[, "index": "analyzed"]   | "type": "text", "index": true    |
```
aertmann: [

Documentation/ElasticMapping-5.x.md
```diff
@@ -0,0 +1,32 @@
...
+The `string` type was used for different things, namely analyzed content to
+be searched as fulltext, as well as in a non-analyzed way for keyword searches.
+
+Thus `string` has been split into the new types `text` and `keyword`. Here is a
+"conversion table":
+
+| 2.x                                       | 5.x                              |
+|-------------------------------------------|----------------------------------|
+| "type": "string", "index": "no"           | "type": "text", "index": false   |
+| "type": "string"[, "index": "analyzed"]   | "type": "text", "index": true    |
```
aertmann: [

Documentation/ElasticMapping-5.x.md
```diff
@@ -0,0 +1,32 @@
...
+## Conflicting field types
+
+Something that has been enforced since version 2 of Elasticsearch is the fact
+that fields of the same name must have the same type within an index, across
+types. See https://www.elastic.co/blog/great-mapping-refactoring#conflicting-mappings.
+
+With the split of `string` into two types, new conflicts of that type may arise.
+
+To fix those conflicts, you may need to rename properties or adjust the mapping
+in your node types.
```
aertmann: out of scope of this change, but thought it might make sense to have a custom property like "search.propertyName" to rename a certain field in the index instead of having to modify existing node types

Classes/Driver/Version1/DocumentDriver.php
```diff
@@ -28,14 +28,21 @@
...
+     * @var NodeTypeMappingBuilderInterface
+     */
+    protected $nodeTypeMappingBuilder;
+
+    /**
      * {@inheritdoc}
      */
     public function delete(NodeInterface $node, $identifier)
     {
+        $nodeTypeMappingBuilder = $this->nodeTypeMappingBuilder;
```
daniellienert: Why are these local references needed?

README.md
```diff
@@ -43,11 +33,26 @@ Then, ensure to update `<your-elasticsearch>/config/elasticsearch.yml` as explai
...
+## Elastic version support
+
+You can switch the Elastic driver by editing ```Settings.yaml```
+(```Flowpack.ElasticSearch.ContentRepositoryAdaptor.driver.version```) with the following value:
+
+* ```1.x``` to support Elastic 1.2 to 1.7
+* ```2.x``` to support Elastic 2.x
+* ```5.x``` to support Elastic 5.x
+
+_Currently the Driver interfaces as not marked as API, and can be changed to adapt to future needs._
```
daniellienert: .. are not marked ...

Classes/Driver/Version1/DocumentDriver.php
kdambekalns: Because `$this->nodeTypeMappingBuilder::convertNodeTypeNameToMappingName($node->getNodeType())` leads to a syntax error in PHP < 7, see https://3v4l.org/Hf8aj

Classes/Driver/Version5/Mapping/NodeTypeMappingBuilder.php
kdambekalns: Not sure I get that, but in that column, it's always about both, but `analyzed` is the default, so even if not given "it'll be there".

Classes/Driver/Version5/Mapping/NodeTypeMappingBuilder.php
```diff
@@ -0,0 +1,144 @@
...
+                $this->adjustStringTypeMapping($fullMapping);
+                $mapping->setFullMapping($fullMapping);
+            }
+
+            // https://www.elastic.co/guide/en/elasticsearch/reference/5.4/dynamic-templates.html
+            $mapping->addDynamicTemplate('dimensions', [
+                'path_match' => '__dimensionCombinations.*',
+                'match_mapping_type' => 'string',
+                'mapping' => [
+                    'type' => 'text'
```
kdambekalns: This should be `keyword`, since the value is always used for exact matches, it does not need to be analyzed. This leads to a mapping conflict, though (hence this "bugfix" of setting it to `text`), that needs to be solved properly.

kdambekalns:
Ok, so now I

- merged in a PR from @nlx-lars that fixes script compilation on ES5
- resolved conflicts
- fixed two other issues

This leaves us with a "proper solution" for the `text` vs `keyword` on `__dimensionCombinations.language`

nlx-lars:
@kdambekalns while indexing works properly now, querying nodes is still broken: https://travis-ci.org/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/jobs/252293767#L754

This shouldn't be too complicated to change: https://www.elastic.co/guide/en/elasticsearch/reference/5.x/query-dsl-filtered-query.html
I probably won't be able to look into that in the next few days though.

Classes/Driver/Version5/Mapping/NodeTypeMappingBuilder.php
aertmann: I meant the weirdly placed [] characters after the end quotes..

kdambekalns: Oh, well. Just there signify "optional", as in `array explode ( string $delimiter , string $string [, int $limit = PHP_INT_MAX ] )`

kdambekalns:
Un-assigning since I'll be on vacation. Feel free to test, fix the tests and come up with a solution for https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/220#discussion_r126612800 in the meantime :)

Classes/Driver/Version1/DocumentDriver.php
daniellienert: Ah, (TIL) - so they could be removed in master (which requires 7.0 lateron

aertmann:
Awesome! Thank you guys!

PR #221, opened 2017-06-28
TASK: Two small tweaks to the `NodeIndexer`

kdambekalns:
- adjusts a parameter comment to be more clear
- removes the use of two extraneous method parameters


kdambekalns:
Rather simple things I saw "along the way".

PR #223, opened 2017-07-08
Allow DimensionCombination Presets to define what's to index

sbruggmann:
As a Feature-Replacement for #206

If you have a lot of DimensionCombinations to structure your content, but you need only a few of them in the index.. then you can save a lot of time with this limitation.


sbruggmann:
This works, but is not enough.
And the configuration complexity is too high, so that a switch to https://github.com/ttreeagency/Flowpack.ElasticSearch.ContentRepositoryQueueIndexer should be checked.

PR #227, opened 2017-07-14
FEATURE: Create one index per dimensions combination

dfeyer:


dfeyer:
So the index work, and trying to query the index with Sense (raw query), the data in the index are correct. So let's try to build a search from in Neos and see if it work.

If you can test the indexing parts, and comments the code, your are welcome some parts feel a bit hacky

gerhard-boden:
Tests fail for ES2. Is that unrelated, or do we need to adapt the tests?

dfeyer:
> Tests fail for ES2. Is that unrelated, or do we need to adapt the tests?

@gerhard-boden Early WIP, i think the package need a big refactoring to support this feature, so currently I don't really care about test, exploration mode ;)

kdambekalns:
Ok, the changes seem relatively small… so what is the status, @dfeyer? Can you outline what is missing and/or needs to be changed to make it work? What is it you are exploring?

dfeyer:
@kdambekalns Currently under heavy test ;)

Basically I m not API with the API, because of the singleton with state, ... let's discuss this later

dfeyer:
OK look like it's ready for first session of review. The API is less than perfect, and it's easy to break the feature (like no dimensions set when something is pushed the ES ... but I think we can not make it better with the current version of the package)

After a full reindex you should have all the alias in place (like this https://gist.github.com/dfeyer/ae72b2b9cdf8eaddc819a1c65807d3b0) so the change is basically transparent and backward compatible, people can continue to query the main alias (typocr, but not this alias point to X indexes). 

Please, review, comments, ... I continue to test this on our current project, but i think we need to test this on more than one project.

My current belief .... the package start to small bad in term of code complexity, maybe parts of the code take some responsability to get the index names, ... So I think we need a refactor of some parts of the package for the next major. That can be a subject for Flowpack Sprint ;)

Anyway I think this change make sense, now let's make it good ;)

dfeyer:
OK the update of the global alias don't work, let's fix this.

An other think about this change, not sure if it's doable but currently the index name is like prefix-context-suffix

- prefix is the index name like configured in Neos.ContentRepository.Search.elasticSearch.indexName
- context is currently the dimensions combination hash
- suffix the timestamp

So currently the context is coupled to the dimensions, but we have plan to support more in the context (Site, Types, ...) so if we can extract this to a dedicated service like IndexingContext based on a an interface, I think we can implement the next step more nicely, but first let's make this PR fully stable.

Classes/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -648,6 +648,8 @@ public function highlight($fragmentSize, $fragmentCount = null)
      */
     public function query(NodeInterface $contextNode)
     {
+        $this->elasticSearchClient->setDimensions($contextNode->getContext()->getTargetDimensions());
```
dfeyer: Maybe we can have a settings or at least a method to query the current specific indexe (current dimensions) or the global index (all dimensions). Some client can found useful to have a query in a foreign language that return results if the document is also available in the current dimensions.

dfeyer:
Area that need testing:

- [ ] Project without dimensions
- [ ] Project based only on the ContentRepository (without Neos)

dfeyer:
In sync with master, ready for review

dfeyer:
I will rebase during the sprint and try to have more reviewers ;)

dfeyer:
Test are broken because the index name change ;) ... I will adapt the test, it's not breaking for existing project, because the old index name is a symlink, but exist only at the end of the indexing process.

Classes/ElasticSearchClient.php
```diff
@@ -28,15 +28,76 @@
 class ElasticSearchClient extends \Flowpack\ElasticSearch\Domain\Model\Client
 {
     /**
+<<<<<<< HEAD
```
daniellienert: Forgotten MergeHead

Classes/ElasticSearchClient.php
```diff
@@ -28,15 +28,76 @@
...
+    {
+        $dimensionValues = $dimensionValues === null ? [] : $dimensionValues;
+        if ($dimensionValues === []) {
+            $this->dimensions = [];
+            $this->dimensionsHash = null;
+            return;
+        }
+        $this->dimensionsHash = $this->dimensionsService->hash($dimensionValues);
+    }
+
```
daniellienert: The called  method getIndexNamePrefix is missing, I guess it got lost with a merge.

Classes/Indexer/NodeIndexer.php
```diff
@@ -257,6 +273,20 @@ public function indexNode(NodeInterface $node, $targetWorkspaceName = null)
             $handleNode($node, $context);
         }
     }
+    
+    protected function toBulkRequest(NodeInterface $node, array $tuple = null)
```
daniellienert: Needs a method comment.

Classes/Indexer/NodeIndexer.php
```diff
@@ -401,6 +449,72 @@ public function updateIndexAlias()
...
+     * Update the main alias to allow to query all indices at once
+     */
+    public function updateMainAlias()
+    {
+        $aliasActions = [];
+        $aliasNamePrefix = $this->searchClient->getIndexNamePrefix(); // The alias name is the unprefixed index name
+
+        $indexNames = $this->indexDriver->indexesByPrefix($aliasNamePrefix);
+        $indexNames = \array_values(\array_filter($indexNames, function ($indexName) {
+            $suffix = '-' . $this->indexNamePostfix;
```
daniellienert: Suffix and postfix, I think we should use only one term

Classes/Indexer/NodeIndexer.php
```diff
@@ -401,6 +449,72 @@ public function updateIndexAlias()
...
+            $indexNameParts = \explode('-', $indexName);
+            return substr($indexName, 0 - strlen($suffix)) === $suffix && count($indexNameParts) === 3;
```
daniellienert: If we already have the indexNameParts, we could use a part instead of strlen calculations

Classes/Service/DimensionsService.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Flowpack\ElasticSearch\ContentRepositoryAdaptor\Service;
+
+/*                                                                                                  *
+ * This script belongs to the TYPO3 Flow package "Flowpack.ElasticSearch.ContentRepositoryAdaptor". *
+ *                                                                                                  *
+ * It is free software; you can redistribute it and/or modify it under                              *
+ * the terms of the GNU Lesser General Public License, either version 3                             *
+ *  of the License, or (at your option) any later version.                                          *
+ *                                                                                                  *
+ * The TYPO3 project - inspiring people to share!                                                   *
```
daniellienert: Wrong Header

dfeyer:
I will update this PR as soon as the current project is in production. 

Main issue currently we use two unmerged PRs (sub command in nodeinex:buid + one index per dimension preset) so we need to end up with two clean PR

kdambekalns:
@dfeyer Did you close this to replace it with something new?

kdambekalns:
Ah, replaced by #245 - awesome!

PR #228, opened 2017-07-15
BUGFIX: Fulltext query is broken

### LINKED ISSUE #197: Slash breaks search query
Searching for a Slash (e.G. "11/2011" or "a/b" ) breaks the ES query.
**Example response**: 
```
                    [reason] => Array
                        (
                            [type] => query_parsing_exception
                            [reason] => Failed to parse query [11/2010]
                            [index] => production-1485418788
                            [line] => 1
                            [col] => 100
                            [caused_by] => Array
                                (
                                    [type] => parse_exception
                                    [reason] => Cannot parse '11/2010': Lexical error at line 1, column 8.  Encountered: <EOF> after : "/2010"
                                    [caused_by] => Array
                                        (
                                            [type] => token_mgr_error
                                            [reason] => Lexical error at line 1, column 8.  Encountered: <EOF> after : "/2010"
                                        )

                                )

                        )
```

ES-Version: 2.4.x
CRA-Version: 3.0.0

@daniellienert can confirm this issue

dfeyer:
The fix introduced to escape the query (see #198) also adds quotes around, so it's not
a full text query anymore.

This change trims the output of `json_encode()` to remove the quotes again.

Fixes #197

### DIFF
--- a/Classes/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Eel/ElasticSearchQueryBuilder.php
@@ -616,7 +616,7 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
     public function fulltext($searchWord)
     {
         // We automatically enable result highlighting when doing fulltext searches. It is up to the user to use this information or not use it.
-        $this->request->fulltext(json_encode($searchWord));
+        $this->request->fulltext(trim(json_encode($searchWord), '"'));
         $this->request->highlight(150, 2);
 
         return $this;


dfeyer:
Maybe we should throw an exception here is the searchTerm is not a string or at least not a scalar type ?

kdambekalns:
If it cannot be transformed into a string, why not…

fdsis:
If searched for `"test"` it gets escaped and then trimmed to `\"test\` which leads to an error:
```
[type] => query_shard_exception
                            [reason] => Failed to parse query [\"test\]
                            [caused_by] => Array
                                (
                                    [type] => parse_exception
                                    [reason] => Cannot parse '\"test\': Lexical error at line 1, column 17.  Encountered: <EOF> after : ""
                                    [caused_by] => Array
                                        (
                                            [type] => token_mgr_error
                                            [reason] => Lexical error at line 1, column 17.  Encountered: <EOF> after : ""
                                        )

                                )
```

It can be patched with: 
```patch
Index: Classes/Eel/ElasticSearchQueryBuilder.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/Classes/Eel/ElasticSearchQueryBuilder.php b/Classes/Eel/ElasticSearchQueryBuilder.php
--- a/Classes/Eel/ElasticSearchQueryBuilder.php	
+++ b/Classes/Eel/ElasticSearchQueryBuilder.php	(date 1649436183235)
@@ -678,7 +678,7 @@
     public function fulltext(string $searchWord, array $options = []): QueryBuilderInterface
     {
         // We automatically enable result highlighting when doing fulltext searches. It is up to the user to use this information or not use it.
-        $this->request->fulltext(trim(json_encode($searchWord, JSON_UNESCAPED_UNICODE), '"'), $options);
+        $this->request->fulltext(trim(json_encode($searchWord, JSON_UNESCAPED_UNICODE), '\\"'), $options);
         $this->request->highlight(150, 2);
 
         return $this;
```

PR #233, opened 2017-08-29
BUGFIX: Fix highlighting with __fulltext* as query_string.fields

akappler:
Elasticsearch highlighting need proper field name match

This adds the necessary field match only to the fulltext() function so that the highlighting works like expected

### DIFF
--- a/Classes/Driver/Version1/Query/FilteredQuery.php
+++ b/Classes/Driver/Version1/Query/FilteredQuery.php
@@ -57,7 +57,8 @@ class FilteredQuery extends AbstractQuery
     {
         $this->appendAtPath('query.filtered.query.bool.must', [
             'query_string' => [
-                'query' => $searchWord
+                'query' => $searchWord,
+                'fields' => ['__fulltext*']
             ]
         ]);
     }


mficzel:
A side effect of this solution is that you will not find any results that are not part of the __fulltext like keyword fields etc.

dfeyer:
@akappler Highlight has been working nice until now (not tested recently), did you have a pointer to the ES documentation that explain what change in recent release ?

akappler:
@dfeyer Sorry, I can't find it in the documentation of elasticsearch at the moment. 

The issue is, that the highlighting function needs a mapping to the fields which should be highlighted. As far as I can remember, query_string works different from query match and therefore the mapping. 

Another solution would be to set `"require_field_match": false` to `true` in the highlighting part see: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-request-highlighting.html#field-match

I think this would fix also the issue @mficzel had.

kdambekalns:
I can confirm highlighting is not working with Elasticsearch 2.4 right now. And looking at the Elasticsearch documentation nothing changed regarding highlighting between 1.7 and 2.4.

The main point is:

> In order to perform highlighting, the actual content of the field is required. If the field in question is stored (has store set to true in the mapping) it will be used, otherwise, the actual _source will be loaded and the relevant field will be extracted from it.
> The _all field cannot be extracted from _source, so it can only be used for highlighting if it mapped to have store set to true.

(see https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-request-highlighting.html)

----

Now, the change @akappler suggested makes highlighting work, but changes the query result (the score changes, not sure if more changes if not using the demo site for testing).

My verdict: **Nope**

----

The suggestion to use `"require_field_match": false` is fine, but… for me suddenly 2 snippets on `__fulltext.text` were returned, and one had a word highlighted, that was not in the query.

My verdict: **Nope**

----

So, let's try the already mentioned `_all` field again, it is what the query is working on:

> The _all field cannot be extracted from _source, so it can only be used for highlighting if it mapped to have store set to true.

This:

    'Neos.Neos:Node':
      search:
        elasticSearchMapping:
          _all:
            store: true

 followed by a `nodeindex:build`, makes it work with:

            $this->request['highlight'] = [
                'fields' => [
                    '_all' => [
                        'fragment_size' => $fragmentSize,
                        'no_match_size' => $fragmentSize,
                        'number_of_fragments' => $fragmentCount
                    ]
                ]
            ];

Downside: for the Neos demo site the index size goes up from 7.65MiB to 11.5MiB, roughly 150%.

My verdict: **Maybe**

----

One more option is the use of a `highlight_query`. That can target `__fulltext*` without influencing scoring. Here's the needed code:

    public function highlight($fragmentSize, $fragmentCount = null)
    {
        if ($fragmentSize === false) {
            // Highlighting is disabled.
            unset($this->request['highlight']);
        } else {
            $highlightQuery = $this->request['query'];
            $highlightQuery['filtered']['query']['bool']['must'][1]['query_string']['fields'] = ['__fulltext*'];

            $this->request['highlight'] = [
                'fields' => [
                    '__fulltext*' => [
                        'fragment_size' => $fragmentSize,
                        'no_match_size' => $fragmentSize,
                        'number_of_fragments' => $fragmentCount,
                        'highlight_query' => $highlightQuery
                    ]
                ]
            ];
        }
    }

This works without changes to the mapping or the "actual query". The downside of the above: The use of a numeric index when adding the field seems to be error prone… That (sh|c)ould be improved.

My verdict: **Probably**

mficzel:
@kdambekalns i found this in the list of breaking changes for elastic 6 https://www.elastic.co/guide/en/elasticsearch/reference/6.x/breaking_60_mappings_changes.html#_the_literal__all_literal_meta_field_is_now_disabled_by_default . Maybe that rules out the penultimate suggestion, at least it makes it harder to support elastic 6 one day.

kdambekalns:
Well, given that we don't even support v5 I won't worry too much about v6. And since we do rely heavily on the `_all` field right now, it doesn't make a difference.

For v6 we need to rework more than that anyway, see https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html

kdambekalns:
I switched to using v5 of Elasticsearch in the meantime, and with that the solution I outlined above does not work (or I was too stupid). Either way, it works fine using master of this package as of today, due to https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/263 being in it.

Given the fact that Elasticsearch 1.x and 2.x are _really_ old by now, I'd suggest we leave this unfixed and recommend an update instead.

PR #235, opened 2017-10-06
FEATURE: IndexName Strategy

dfeyer:
Currently the only way to define the Index name, is by Settings. With
this new Interface, it's now possible to implement other Strategies.

### DIFF
--- a/Classes/ElasticSearchClient.php
+++ b/Classes/ElasticSearchClient.php
@@ -11,6 +11,7 @@ namespace Flowpack\ElasticSearch\ContentRepositoryAdaptor;
  * source code.
  */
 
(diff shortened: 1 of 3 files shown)


dfeyer:
Not sure if it's part of the PR, i think no, but we need to provide some kind of support for this https://github.com/Flowpack/Flowpack.SearchPlugin/blob/master/Configuration/Settings.yaml#L23

Index configuration by name, should support index name placeholder (filled by the strategy). So by defautl everything work (the current change is not breaking, until you use a custom strategy)

Now the big question, this strategy belong to the ContentRepositoryAdapter package ? or the ElasticSearch package ... i think the ContentRepositoryAdapter is the correct place, because no convention in the ElasticSearch about index naming.

@kdambekalns What do you think ?

PR #236, opened 2017-11-24
TASK: Catch query errors and return empty result

kdambekalns:
Instead of simply ignoring an exception during query execution the
`ElasticSearchQueryBuilder` will now log an exception during `fetch()`
and return an empty result.

### DIFF
--- a/Classes/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Eel/ElasticSearchQueryBuilder.php
@@ -549,20 +549,25 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
      */
     public function fetch()
     {
-        $timeBefore = microtime(true);
-        $request = $this->request->getRequestAsJson();
-        $response = $this->elasticSearchClient->getIndex()->request('GET', '/_search', [], $request);
-        $timeAfterwards = microtime(true);
+        try {
+            $timeBefore = microtime(true);
+            $request = $this->request->getRequestAsJson();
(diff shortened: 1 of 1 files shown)


Classes/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -549,20 +549,25 @@ public function getFullElasticSearchHitForNode(NodeInterface $node)
...
+            $this->result['nodes'] = [];
+            if ($this->logThisQuery === true) {
+                $this->logger->log(sprintf('Query Log (%s): %s -- execution time: %s ms -- Limit: %s -- Number of results returned: %s -- Total Results: %s',
+                    $this->logMessage, $request, (($timeAfterwards - $timeBefore) * 1000), $this->limit, count($this->result['hits']['hits']), $this->result['hits']['total']), LOG_DEBUG);
+            }
+            if (array_key_exists('hits', $this->result) && is_array($this->result['hits']) && count($this->result['hits']) > 0) {
+                $this->result['nodes'] = $this->convertHitsToNodes($this->result['hits']);
+            }
+        } catch (ApiException $exception) {
+            $this->logger->log('Error while fetching results: ' . $exception->getMessage(), LOG_ERR);
```
dfeyer: As this is an Exception, please use logException here, some people use sentry to manage exception, and it's nice to have all the details.

kdambekalns: Fair enough, didn't think of that. I checked the log, and the result looks the same (since the full details from ES are in the message itself.)

PR #242, opened 2018-02-07
BUGFIX: Change TYPO3CR to ContentRepository in settings example

kdambekalns:


kdambekalns:
Dang, #244 is indeed better. :)

PR #244, opened 2018-02-08
BUGFIX: Change TYPO3CR to ContentRepository in settings example

dlubitz:


dlubitz:
Related and half duplicate: #242

PR #251, opened 2018-04-27
TASK: Adjust version constraints for Neos 4

mficzel:


daniellienert:
@mficzel I am currently testing the elastic 5.x driver in real world projects. There some things to optimize. So it would be a good idea to branch from
Master before the es 5 driver merge and update the constraints there. What do you think

mficzel:
@daniellienert in that case we should create a 4.x branch for the adapter with the last version before the es5 merge. I do not have the access-right to do this but if you do adjusting the target branch of this pr should be easy.

PR #253, opened 2018-05-01
TASK: Use source for inline scripts as inline is deprecated

daniellienert:


daniellienert:
@kdambekalns the 5.6 documentation already uses `source` - https://www.elastic.co/guide/en/elasticsearch/reference/5.6/modules-scripting-using.html I didn't find a deprecation warning in the docs, but on my ES 5.6 instance my log is spammed with 

`Specifying script source with [inline] is deprecated, use [source] instead`

daniellienert:
@kdambekalns: `Conclusion: don't install elasticsearch < 5.6` :)
Ok, I agree.

PR #252, opened 2018-05-01
TASK: Migrate include_in_all to copy_to for ES 5.x

daniellienert:
Include_in_all is deprecated with elasticsearch 5.x and raises warnings on index creation

PR #254, opened 2018-05-01
BUGFIX: Also map elasticsearch config that is passed directly

daniellienert:
adjustStringTypeMapping() is also called directly with `$propertyConfiguration['search']['elasticSearchMapping‘]` - in this case the fields are not altered. This raises a lot of elasticsearch deprecation warnings.


kdambekalns:
@daniellienert This is effectively included in #252, no?

Classes/Driver/Version5/Mapping/NodeTypeMappingBuilder.php
```diff
@@ -140,6 +136,16 @@ protected function adjustStringTypeMapping(array &$mapping)
                     $item['index'] = true;
                 }
             }
+        };
+
+        $adjustStringTypeMapping($mapping);
+
+        foreach ($mapping as &$item) {
+            if (!is_array($item)) {
+                continue;
```
kdambekalns: I'd remove the ngetaion and move the calls into the loop.

PR #255, opened 2018-05-02
BUGFIX: Fix highlighting by searching in all fulltext fields

### LINKED ISSUE #232: Bug: Highlighting doesn't add em-tags
Hi,
I had some trouble with getting the search word highlighted in the result.
The problem with the current version is, that elasticsearch doesn't think that the fields definition in the highlight section of the search query doesn't match with the searched fields. Therefore the highlighting doesn't work and changes via `.highlight()` eel helper have no effects. 

My suggested solution would be to add the missing parameters `fields => ['__fulltext*']` in the FilteredQuery.php at the `fulltext` function.

After I added this, the search word got highlighted as expected.

A quick workaround is to tweak the fusion searchQuery in your own package like this:

```
prototype(Flowpack.SearchPlugin:Search) < prototype(Neos.Neos:Content) {

    queryStringFields = Neos.Fusion:RawArray {
        fields = Neos.Fusion:RawArray {
            0 = "__fulltext*"
        }
    }

    searchQuery = ${this.searchTerm ? Search.query(site).fulltext(this.searchTerm).request('query.filtered.query.bool.must.1.query_string', this.queryStringFields).nodeType('Neos.Neos:Document') : null}    
}
```

I'll add an PR.

daniellienert:
Only fields are highlighted which are searched on.
This changes the search from _all to the fullText fields.

Fix #232

### DIFF
--- a/Classes/Driver/Version2/Query/FilteredQuery.php
+++ b/Classes/Driver/Version2/Query/FilteredQuery.php
@@ -18,4 +18,16 @@ use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\Version1;
  */
 class FilteredQuery extends Version1\Query\FilteredQuery
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function fulltext($searchWord)
+    {
+        $this->appendAtPath('query.filtered.query.bool.must', [
+            'query_string' => [
+                'query' => $searchWord,
+                'fields' => ['__fulltext*']
+            ]
+        ]);
+    }
 }
--- a/Classes/Driver/Version5/Query/FilteredQuery.php
+++ b/Classes/Driver/Version5/Query/FilteredQuery.php
@@ -27,7 +27,8 @@ class FilteredQuery extends Version1\Query\FilteredQuery
     {
         $this->appendAtPath('query.bool.must', [
             'query_string' => [
-                'query' => $searchWord
+                'query' => $searchWord,
+                'fields' => ['__fulltext*']
             ]
         ]);
     }
--- a/Configuration/NodeTypes.yaml
+++ b/Configuration/NodeTypes.yaml
@@ -140,30 +140,37 @@
             'h1':
               type: string
               include_in_all: true
+              index: analyzed
               boost: 20
             'h2':
               type: string
               include_in_all: true
+              index: analyzed
               boost: 12
             'h3':
               type: string
               include_in_all: true
+              index: analyzed
               boost: 10
             'h4':
               type: string
               include_in_all: true
+              index: analyzed
               boost: 5
             'h5':
               type: string
               include_in_all: true
+              index: analyzed
               boost: 3
             'h6':
               type: string
               include_in_all: true
+              index: analyzed
               boost: 2
             'text':
               type: string
               include_in_all: true
+              index: analyzed
               boost: 1
     '_hiddenInIndex':
       search:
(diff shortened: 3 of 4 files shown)


daniellienert:
Hey @kdambekalns / @skurfuerst  / @dfeyer, 
in order to fix the highlighting in ES > 2 we need to search on the fields where highlighting should be applied afterwards.
When I do search on all __fulltext.* fields, then we could skip the copying everything to _all - am I right?
So we could remove the include_in_all parameter with the next version?

Classes/Driver/Version2/Query/FilteredQuery.php
```diff
@@ -18,4 +18,16 @@
...
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function fulltext($searchWord)
+    {
+        $this->appendAtPath('query.filtered.query.bool.must', [
+            'query_string' => [
+                'query' => $searchWord,
+                'fields' => ['__fulltext*']
```
dfeyer: This is quiet breaking and hard to detect for existing project, maybe `['_all', '__fulltext*']` is less breaking, should just improve the score of the matching string in `_all` and `__fulltext*`

dfeyer: Not sure if highlights work in this condition, but it should

dfeyer:
In think it's a badly breaking changes. Many people use the _all so if we can fix the highlights without breaking the current behavior badly I all in favor ;)

Classes/Driver/Version2/Query/FilteredQuery.php
daniellienert: It would work.

daniellienert:
Hey @dfeyer,
thanks for the feedback. Thats why I'd like to discuss that. 
My arguments for removing the include_in_all configuration are:
- using the _all field for search in production is a bad practice and deprecated and makes the request slower
- Removing it, would clean up the configuration. Using `include_in_all` is not recommended - we have the `search.fulltextExtractor` where we support boosting
- The next version could be a major and could be breaking
- We could deprecate include_in_all and automatically index all fields with this setting into the __fullTextParts.text field

kdambekalns:
There is a pending PR to fix this in #233 as well. Wisdom from both should be combined, I'll give that a try. :)

kdambekalns:
So, after reading up on the comments here and in #233 my conclusion would be: use this as a solution to get rid of using the `_all` field, but don't disguise it as a bugfix for that… My suggestion would be:

- Fix highlighting using a `highlight_query` like shown in #233 - this avoids any change to scoring or the `_all` field in current versions.
- In the master branch, switch to using `__fulltext*` and get rid of `_all` completely (based on this PR).

What do you think @daniellienert?

daniellienert:
Hey @kdambekalns,
mhm after reading comments and change of #233:

- This as well as #233 changes the behavior and does not search in the `_all`field. We need to set fields to ['_all', '__fulltext*'] to keep the previous behavior and justify a bugfix.
- According to my tests and the functional tests included, the fulltext fields need to configured as analyzed for the ES5 driver to work.

So I would change this one to also search in _all and then remove all the include_in_all and _all stuff (:p) in a breaking change for the upcoming major.

kdambekalns:
In #233 I added a comment that uses a `highlight_query` that is only used for highlighting. That does not affect the search itself at all. The PR itself as it is currently does, true.

When it comes to Elasticsearch 5.x, I think some divergence from the 1.x and 2.x behaviour even now is acceptable.

daniellienert:
Closing this pull request and rebuilding it onto master

PR #262, opened 2018-07-04
BUGFIX: Fix highlighting using a special highlight query

daniellienert:


daniellienert:
Hey @kdambekalns, I tried to fix the highlighting using the code you suggested in https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/233. Sadly - according to my tests - this does not work. Could you please have a look?

PR #276, opened 2018-08-13
TASK: Configure ESQueryBuilder as implementation of the QueryBuilderInterface

daniellienert:
When you try to extend the ElasticSearchQueryBuilder and with this adding a second class implementing
the QueryBuilderInterface, the ObjectManager is not able to automatically select the implementing class
any more. This adds this missing configuration


bwaidelich:
Not sure why I was suggested as reviewer for this, but it looks good to me by reading :)

daniellienert:
Thanks @bwaidelich, I was sure you now how an object configuration hast to look like :p

PR #278, opened 2018-08-26
TASK: Expose cache flag on es query

johannessteu:


johannessteu:
We'll it's not just Fusion. We do use the QueryBuilder in PHP Code for instance as well. And depending on your use case it's necessary to fetch a query, do some mutations an refetch but you do not get the new result beacuse the cache kicks in. It also has some drawbacks in Testing as well if you run a query, add some stuff, run a persistAll and refetch.

PR #279, opened 2018-09-09
FEATURE: More like this

daniellienert:
This implements an convinent API to build more like this queries.
For details see the [Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-mlt-query.html)


Classes/Driver/AbstractQuery.php
```diff
@@ -119,7 +135,6 @@ public function aggregation($name, array $aggregationDefinition, $parentPath = '
      * @param string $parentPath The parent path to add the sub aggregation to
      * @param string $name The name to identify the resulting aggregation
      * @param array $aggregationConfiguration
-     * @return QueryInterface
```
kdambekalns: So it should be `@ return void`now?

dfeyer: For those little nitpick, I think we can use the github edit feature, add the missing bit and done ;)

PR #280, opened 2018-09-10
TASK: Add a new `executeUncached()` method to ElasticSearchQueryBuilder

johannessteu:
This change introduces a new method `executeUncahced` to the ElasticSerachQueryBuilder. This method will actually execute the ElasticSearchQuery but the result will be uncached.


johannessteu:
As a alternative to #278

dfeyer:
I don't see this one before merging the #278 :) Revert or not ?

daniellienert:
@dfeyer, I think this one is the cleaner approach, so I would say yes.

PR #284, opened 2018-10-29
TASK: Add important note for the suggestion on 5.x

markusguenther:
I think it is important to know,  that when using Elasticsearch 5.x the suggest and autocomplete feature is actually not working.

I also only read the docs before using the package and thought awesome that this is there.
But to prevent more slack messages on that topic, I think a mention in the docs is worth it until it is fixed/reimplemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ _Currently the Driver interfaces are not marked as API, and can be changed to ad
 **Note:** When using Elasticsearch 5.x changes to the types may need to be done in your mapping.
 More information on the [mapping in ElasticSearch 5.x](Documentation/ElasticMapping-5.x.md).
 
+**IMPORTATN:** When using Elasticsearch 5.x the suggest and autocomplete feature is actually not working.
+
 ### Elasticsearch Configuration file elasticsearch.yml
 
 There is a need, depending on your version of Elasticsearch, to add specific configuration to your
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ _Currently the Driver interfaces are not marked as API, and can be changed to ad
 **Note:** When using Elasticsearch 5.x changes to the types may need to be done in your mapping.
 More information on the [mapping in ElasticSearch 5.x](Documentation/ElasticMapping-5.x.md).
 
-**IMPORTATN:** When using Elasticsearch 5.x the suggest and autocomplete feature is actually not working.
+**IMPORTANT:** When using Elasticsearch 5.x the suggest and autocomplete feature is actually not working.
 
 ### Elasticsearch Configuration file elasticsearch.yml
 


kdambekalns:
@daniellienert Is that true? Do you happen to know?

Also, what does it take to fix?

markusguenther:
@kdambekalns You can make the suggestions work with the Flowpack.searchplugin when @aertmann patches will be merged. But until this merges the suggest is not working.

daniellienert:
Damn, that was one of the task that I planned to take care of during the sprint.
I think it is not a good idea to add this hint to the master branch of this package, as master only works with ES5. We need to merge @aertmann's branch and release both packages with according versions.

markusguenther:
I am fine with that. Just added the note to prevent that others have the same issue.

Because I also read the docs an thought it works ;)

kdambekalns:
What is the status here, @daniellienert & @markusguenther?

daniellienert:
I still think it is confusing to add a line about “when using es 5” when using es 5 is the only option.

markusguenther:
Then lets integrate @aertmann commits in his forks and then the search plugin is compatible and we don´t need this hint.

https://github.com/aertmann/Flowpack.SearchPlugin/commits/master

dfeyer:
I work on a big change so if something big need to be merged on master (I don't know if @aertmann branch cited by @daniellienert include big changes), please keep me in sync 

See https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/298

dfeyer:
An other question, it's not clear for me. Did the suggestion are broken in the ContentRepositoryAdapter ... or in the SearchPlugin, who need to be adapted for recent change in the ContentRepositoryAdapter ?

If the suggestion feature are broken in the ContentRepositoryAdapter it's not an acceptable situation. It's a feature, that is part of the package since some version now, why did we release version (or even master) without the complete feature set ?

markusguenther:
@dfeyer It is just the searchplugin. Aske fixed it on the 13th of september in his fork and it works like charm. Don´t know if something was missing. For me that solved the issue itself.

https://github.com/aertmann/Flowpack.SearchPlugin/commits/master

markusguenther:
From my point of view we just need to fix the search plugin with @aertmann four commits and release a new version. Then this PR would be obsolete.

kdambekalns:
> @dfeyer It is just the searchplugin. Aske fixed it on the 13th of september in his fork and it works like charm. 

That is exactly PR https://github.com/Flowpack/Flowpack.SearchPlugin/pull/49 – I just asked @daniellienert if he knows the current state of affairs. I completely lost track…

daniellienert:
Finally tested and merged Flowpack/Flowpack.SearchPlugin#49 and released https://github.com/Flowpack/Flowpack.SearchPlugin/releases/tag/4.1.0! That means this PR is not necessary anymore.

Issue #291, opened 2019-01-29
4.1.1 broken (fatal error)


aertmann:
The upmerge from 3.0 to 4.0 caused a fatal error

    PHP Fatal error:  Declaration of Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\Version1\Query\FilteredQuery_Original::fulltext($searchWord) must be compatible with Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\QueryInterface::fulltext(string $searchWord, array $options = Array) in Data/Temporary/Production/SubContextAws/Cache/Code/Flow_Object_Classes/Flowpack_ElasticSearch_ContentRepositoryAdaptor_Driver_Version1_Query_FilteredQuery.php on line 20

Most likely caused by #290, however not visible as it was a clean merge.

Before https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/4.1.0/Classes/Driver/Version1/Query/FilteredQuery.php#L56
After https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/4.1.1/Classes/Driver/Version1/Query/FilteredQuery.php#L56

PR #292, opened 2019-01-31
BUGFIX: fix upmerge gone wrong

dimaip:
I still struggle to understand how it happened.
As I understand, it had to be so that changes in 3.0 branch had to be more recent than in 4.0. But how then would you explain things like reappearing of `TYPO3CR` in the readme?!

Fixes: #291

### DIFF
--- a/Classes/Driver/QueryInterface.php
+++ b/Classes/Driver/QueryInterface.php
@@ -69,13 +69,14 @@ interface QueryInterface
     public function from($size);
 
     /**
-     * Match the searchword against the fulltext index
+     * Match the search word against the fulltext index
      *
      * @param string $searchWord
+     * @param array $options Options to configure the query_string
      * @return void
      * @api
      */
-    public function fulltext($searchWord);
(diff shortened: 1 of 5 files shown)


dimaip:
@kdambekalns do you have any ideas of how it could happen?

kdambekalns:
No upmerge gone wrong…

The story is quite different. The code you fixed was added in c9461085b948526f786ca41d5675de380201f55f to the _master_ branch. 4.1.0 was created on 4a7f016f0bd1c78f62fe3d6009b2426e59150ece on the _master_ branch… But 4.1.1 was created off the _4.0_ branch (which is quite correct), thus it never contained those changes. As such it was indeed broken.

The 4.0 branch needs 4a7f016f0bd1c78f62fe3d6009b2426e59150ece cherry-picked (which you indirectly already did with your fix, I guess) and then everything is in sync again.

dimaip:
@kdambekalns thanks for explaining!

Issue #300, opened 2019-05-08
Change NodeType results in error


simonedelmann:
I just installed this package and set up ElasticSearch for a simple site package, I'm just creating.

The bug: Changing a node's type immediately results in an error.
This appears for changing document aswell as content nodes.

> 
> Elasticsearch request failed. 
> [GET http://search:9200//_search/scroll?scroll=1m]: Array ( [root_cause] => Array ( [0] => Array ( [type] => illegal_argument_exception [reason] => request [//_search/scroll] contains unrecognized parameter: [scroll] ) ) [type] => illegal_argument_exception [reason] => request [//_search/scroll] contains unrecognized parameter: [scroll] ) ; Response body: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"request [//_search/scroll] contains unrecognized parameter: [scroll]"}],"type":"illegal_argument_exception","reason":"request [//_search/scroll] contains unrecognized parameter: [scroll]"},"status":400} Request data: DnF1ZXJ5VGhlbkZldGNoBQAAAAAAAA6AFkxiQjJSS2gxVDZlbUIyUV9YWklPeEEAAAAAAAAOghZMYkIyUktoMVQ2ZW1CMlFfWFpJT3hBAAAAAAAADn8WTGJCMlJLaDFUNmVtQjJRX1haSU94QQAAAAAAAA6BFkxiQjJSS2gxVDZlbUIyUV9YWklPeEEAAAAAAAAOgxZMYkIyUktoMVQ2ZW1CMlFfWFpJT3hB
> 

Edit: I'm using the latest versions of all related packages:
Neos/Neos: 4.3.0
Flowpack/ElasticSearch: 2.0.4
Flowpack/ElasticSearch.ContentRepositoryAdapter: 5.0.1
ElasticSearch is running in a docker container using the official Elasticsearch image, version 5.

bweinzierl:
Had the same issue when switching document nodetypes. This PR fixes the error but it is not tested if the intended behaviour works again. But it should :)
It would be great if someone with a bit more insight into this package could check.
Maybe @daniellienert ?

DrillSergeant:
Just encountered the same issue

bweinzierl:
@DrillSergeant check out my PR :)

DrillSergeant:
[...]

PR #307, opened 2019-06-17
BUGFIX: restructure invalid elasticsearch scroll request

bweinzierl:
This fixes an error thrown when changing the node-type of a document node (see issue #300 ).
To be honest i am not 100% sure what that code in DocumentDriver.php actually does but it seems the request to the elasticsearch instance had two errors:
-  It did not include the indexname (hence the last parameter was "false")
-  It had a wrong structure so elasticsearch 5.6 threw an error

It would be great if someone could look at it and verify if my fix brakes the intended behaviour. I did only test if the error disappeared.


daniellienert:
Fixed with #315

PR #311, opened 2019-06-25
BUGFIX: Use multi_match instead of query_string

gerhard-boden:
By using ``multi_match`` instead of ``query_string`` within our search
query, we prevent the accidental injection of Lucene search query
strings. Currently an exception is thrown when adding ``"`` to your search query.

Using ``multi_match`` instead should lead to the same quality of results
and is less prone to user errors, because in 99% of cases the search is
used for classic search terms and no end user is expected to know the
compact Lucene query string syntax.

see: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-multi-match-query.html
see: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-query-string-query.html
see: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-query-string-query.html#query-string-syntax


Classes/Driver/Version5/Query/FilteredQuery.php
```diff
@@ -61,6 +61,7 @@ public function fulltext(string $searchWord, array $options = []): void
     {
         $this->appendAtPath('query.bool.must', [
             'query_string' => array_merge($options, [
+//            'multi_match' => array_merge($options, [
```
daniellienert: Hey @gerhard-boden - seems something went wrong here :)

gerhard-boden: Oh yes, was testing if my change broke the test or if the tests have been broken all along. But obviously it's my change 🙈 (Didn't want to install elasticsearch locally for just running the test)

Will revert to my change and try to figure out why I'm breaking the tests.

PR #312, opened 2019-08-03
!!! TASK: Refactor and simplify logging for compatibility with Flow 6.0

daniellienert:
- Change the Interface and logger calls
- Remove the specified elasticsearch logger and logfile


kdambekalns:
I like having a separate log file for Elasticsearch, to be honest. Can this be reproduced with the PSR logger?

daniellienert:
Talked to @kdambekalns and her doesn't insist of the separate logger. All log files are tagged with the corresponding package, so a `tail -f | grep ElasticSearch System.log` would return all messages related to Elasticsearch.

PR #298, opened 2019-04-05
FEATURE: Adapt indexing to create one index per dimensions preset

dfeyer:
This change also use sub command to lower to amount of memory needed by a full indexing.

# Features

- [x] Sub command to limit memory usage
- [x] Split bulk request based on size/length limits
- [x] One index per node combination hash

# Todo

- [x] Split indexing across multiple indexes
- [x] Switch index, based on context node target dimensions, during query
- [x] Use sub command during `nodeindex:build` to avoid high memory usage
- [x] Index invisible content
- [x] CLI command to show mapping between index name and dimensions preset
- [x] Add support for per index configuration (analyzer, stemming, ...)
- [ ] Use the old configuration for all indices if no more specific configuration provided (never merge the old and the new config)
- [ ] Update documentation
- [ ] Fix tests and improve test coverage

PR #314, opened 2019-11-18
FEATURE:  One index per dimension preset

daniellienert:
This is a copy of #298 with further improvements.

This change also use sub command to lower to amount of memory needed by a full indexing.

# Features

- [x] Sub command to limit memory usage
- [x] Split bulk request based on size/length limits
- [x] One index per node combination hash

# Todo

- [x] Split indexing across multiple indexes
- [x] Switch index, based on context node target dimensions, during query
- [x] Use sub command during `nodeindex:build` to avoid high memory usage
- [x] Index invisible content
- [x] CLI command to show mapping between index name and dimensions preset
- [x] Add support for per index configuration (analyzer, stemming, ...)
- [x] Use the old configuration for all indices if no more specific configuration provided (never merge the old and the new config)
- [x] Update documentation
- [x] Fix tests and improve test coverage

*Performance Improvements*
- [x] Do not count the batch sizes on every add, store the current size instead

# Performance testing

Done on my 2018 Macbook

## Project: Neos Demo
|Version| Time for `flow nodeindex:build --workspace=live`| Steps|
|--------|--------------------------------------------------|------|
|6.0| 1,85s user 0,89s system 18% cpu 15,147 total|
|#314 @f8721aa|8,74s user 7,63s system 37% cpu 43,921 total| Create 21.22<br>Indexing ... 18.14<br>Refresh 2.68<br>Aliases... 2.89<br>|
|#314 @137a2f2|1,68s user 3,32s system 18% cpu 26,996 total| Create 39.99<br>Indexing ... 0.90<br>Refresh 0.24<br>Aliases... 0.26<br>|

## Project: Customer Project 30K nodes 5 languages
|Version| Time for `flow nodeindex:build --workspace=live`| Steps|
|--------|--------------------------------------------------|------|
|6.0| 102,00s user 10,93s system 17% cpu 10:37,15 total|
|#314 @f8721aa|122,27s user 62,25s system 14% cpu 20:59,09 total|Create 23.05<br>Indexing 1,224.47<br>Refresh 2.76<br>Aliases... 2.25|
|#314 @98dfdd0|107,74s user 21,74s system 13% cpu 16:10,56 total|Create 25.27<br>Indexing 938.00<br>Refresh 2.66<br>Aliases...   2.23|
|#314 @137a2f2|51,89s user 13,33s system 17% cpu 6:04,29 total|Create 27.27<br>Indexing 330.35<br>Refresh 2.38<br>Aliases 2.38<br>|


daniellienert:
I found the problem, why this version is such slow:
1. The NodeIndexCommandController:L262 iterates over all dimension combinations and starts the indexing process.
2. The indexing process traverses the tree and passes all nodes to the nodeIndexer.
3. The NodeIndexer:L290 then iterates again over all dimension combinations and indexes the node in all combinations.

For instances, where the tree is mostly translated, that ends up in complexity of O(nodes x dimensioncombinations^2) where O(nodes x dimensioncombinations) is sufficient.

As the command controller is able to index a dimension in a sub-process. We should handle dimensions there. But as we traverse the tree, we do not get nodes which are translated further down the tree. 

I think its best to skip traversing in favor of just iterating the node table, like the queeindexer is doin it already.

Classes/Indexer/NodeIndexer.php
```diff
@@ -430,6 +537,72 @@ public function updateIndexAlias(): void
...
+    public function updateMainAlias()
+    {
+        $aliasActions = [];
+        $aliasNamePrefix = $this->searchClient->getIndexNamePrefix(); // The alias name is the unprefixed index name
+
+        $indexNames = $this->indexDriver->indexesByPrefix($aliasNamePrefix);
+        $indexNames = \array_values(\array_filter($indexNames, function ($indexName) {
+            $suffix = '-' . $this->indexNamePostfix;
+            $indexNameParts = \explode('-', $indexName);
+            return substr($indexName, 0 - strlen($suffix)) === $suffix && count($indexNameParts) === 3;
```
theilm: If the configured index name itself contains a dash, counting the indexNameParts does not work as expected and the main alias can't be updated.

Classes/ElasticSearchClient.php
```diff
@@ -37,13 +38,94 @@ class ElasticSearchClient extends Client
...
+     * @throws \Exception
+     */
+    public function withDimensions(\Closure $closure, array $dimensionValues = []): void
+    {
+        $previousDimensionHash = $this->dimensionsHash;
+        try {
+            $this->setDimensions($dimensionValues);
+            $closure();
+            $this->dimensionsHash = $previousDimensionHash;
+        } catch (\Exception $exception) {
```
theilm: You could use a finally block that contains the dimensionsHash-Reset; that avoids code duplication and you do not need to re-throw a catched exception.

Classes/Indexer/NodeIndexer.php
daniellienert: @theilm, should be fixed

theilm: @daniellienert I reviewed your commit and tested it, it works now! Thanks! :+1:

Classes/Driver/IndexDriverInterface.php
```diff
@@ -19,13 +19,22 @@
...
+
     /**
      * Get the list of Indexes attached to the given alias
      *
      * @param string $alias
      * @return array
      */
-    public function indexesByAlias(string $alias): array;
+    public function getIndexeNamesByAlias(string $alias): array;
+
```
theilm: I guess it should be `getIndexNamesByAlias` (same for `getIndexNamesByPrefix`? Sorry for the nitpick, but as it's in the Interface...

daniellienert: uuh, fixed, thanks.

kitsunet:
@daniellienert I know I am VERY late to the party, but the commit `80a2e152c117ac90c91deeb82eb177fa3650ee23` seems pretty fishy [TASK: Move contextNode to ElasticSearchClient](https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/314/commits/80a2e152c117ac90c91deeb82eb177fa3650ee23)
The ElasticSearchClient is singleton so creating multiple queries with different context nodes (that may actually have different contexts) is bound to lead to problems as only the last queries context node wins, especially due to "lazy" nature of it all. The context node is not used right away but sometime later when the query is actually executed and nodes are hydrated but by that time the context given to the query might no longer be the one in the client. It's probably one of those cases were it works fine 90% of time, but if it doesn't it's bad. IMHO we should revert this behavior.

daniellienert:
Hey @kitsunet, Uuh, thats bad ... but AFAIS it also wasn't much better before - some parts, like the dimension has been set to the singleton before. The concept to have the QueryBuilder as an interface to fusion which is not Singleton, configuring a Singleton object in the background seems wrong. Is the `ElasticSearchClient` really needed to be a singleton?

kdambekalns:
I tried in my "problem project" with the context node back in the `ElasticSearchQueryBuilder`, but that did not solve my problem(s) with v7 compared to v6:

- nodes with hidden parents returned
- node returned that formerly were not returned

Since that change has been in v6, that might not be a surprise… We left behind flowpack/elasticsearch-contentrepositoryadaptor 6.0.3 in June 2020. So there must be something else behind my problem.

Issue #309, opened 2019-06-25
Avoid underscorce prefix in field name


gerhard-boden:
Maybe we should try to avoid underscores in our field names when doing the mapping, since this raises issues when using Kibana. Also see https://github.com/elastic/kibana/issues/14856.

Not sure how much effort it would take to just avoid the `_` prefixes or if its even worth it for the probably small number of Kibana users.

daniellienert:
Hey @gerhard-boden , I also stumbled upon this recently. As the next major versions of the ES packages are breaking anyways it could be good Idea to change this. We need some prefix nevertheless. What could taht be? 'neos_', 'Internal_'?

PR #319, opened 2020-04-04
WIP: !!! TASK: Rename fields to beats convention

daniellienert:
Previously fields that are not defined via properties have been started with an underscore or sometimes with two underscores. This is generally not a problem for elasticsearch but kibana treats these fields as elasticsearch-internal and doesn’t aloow to analyze them.

Now the fields follow the rules defined in https://www.elastic.co/guide/en/beats/devguide/current/event-conventions.html

The fields are renamed as follows:

| Old | New |
|-----|-----|
|__identifier |neos_node_identifier |
|__parentPath |neos_parent_path |
|__path |neos_path|
|__typeAndSupertypes|neos_type_and_supertypes|
|__workspace |neos_workspace|
|_creationDateTime|neos_creation_date_time|
|_hidden|neos_hidden|
|_hiddenBeforeDateTime |neos_hidden_before_datetime|
|_hiddenAfterDateTime|neos_hidden_after_datetime|
|_hiddenInIndex|neos_hidden_in_index|
|_lastModificationDateTime|neos_last_modification_date_time|
|_lastPublicationDateTime|neos_last_publication_date_time|
|__fulltextParts|neos_fulltext_parts|
|__fulltext|neos_fulltext|

closes: #309

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ Neos:
     '_hiddenBeforeDateTime':
       search:
 
-        # a date should be mapped differently, and in this case we want to use a date format which
+        # A date should be mapped differently, and in this case we want to use a date format which
         # Elasticsearch understands
         elasticSearchMapping:
           type: DateTime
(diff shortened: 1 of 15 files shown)


Configuration/NodeTypes.yaml
```diff
@@ -3,69 +3,64 @@
     fulltext:
       enable: true
   properties:
-    '__identifier':
+    'neos.node_identifier':
```
radmiraal, line 6: is it on purpose you use the dot notation here? I'd assume the snake_case to be used.

README.md
```diff
@@ -260,7 +260,7 @@ Neos:
  # NodeTypes.yaml
 'Neos.Neos:Timable':
   properties:
-    '_hiddenBeforeDateTime':
+    'neos_hidden_before_datetime':
```
radmiraal, line 263: here you replaced the actual node property and not the elasticsearch mapping

radmiraal:
I like the naming convention, was thinking about it today while trying out the master branch of the package and actually had the same naming idea.

Configuration/NodeTypes.yaml
daniellienert: Nope, was an experiment to use a "neos" object with subfields. But that seems to need a serious refactoring of Flowpack.Elasticsearch which I don't want to do this time

daniellienert:
Thanks for the feedback. Polished the PR and deactivated the indexing of the underscore prefixed fields. Will merge when tests are green.
That would be the last breaking change. Lets get this out eventually :)

Issue #323, opened 2020-04-30
Reimplement indexing of asset content


daniellienert:
Since Elasticserach v6 the mapper-attachment plugin is not available anymore. 
It is replaced by the ingest-attachment plugin which requires an ingest pipeline to be configured.

There are currently 2 PRs open with different approaches.

#324 Provides configurable ingest pipelines and uses one to ingest an asset to a predefined fields (See the disadvantages)
#325 Uses an ad-hoc pipeline via the simulate API

I currently prefer to merge #325

PR #325, opened 2020-05-01
FEATURE: Implement asset indexing using attachment-ingest plugin

daniellienert:
This introduces a new interface `AssetExtractorInterface` which handles asset content extraction and a DTO to transfer the data.

The provided implementation uses the ingest-attachment plugin via an ad-hoc introduced pipeline.

The data can be extracted using the eel operation `Indexing.extractAssetContent(value)`. Which extracts content and meta data using a separate elasticsearch call.

Despite of abusing the simulate API, this solution fits way better in the current concepts and needs way less moving parts.

Needs https://github.com/neos/typo3cr-search/pull/36 to be merged first.

Closes: #323


Classes/AssetExtraction/IngestAttachmentAssetExtractor.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    protected $elasticsearchClient;
+
+    /**
+     * @Flow\Inject
+     * @var LoggerInterface
+     */
+    protected $logger;
+
+    public function extract(AssetInterface $asset): AssetContent
```
andrehoffmann30, line 43: Methode comment is missing here

PR #335, opened 2020-05-29
TASK: Exchange scrolling and deleting by simple _delete_by_query

### LINKED ISSUE #333: Exchange the scroll/bulk request query with a simple _delete_by_query in deleteDuplicateDocumentNotMatchingType
Delete by query was not available in ES 2.x which leads to this code, which seems to have bugs. Replace it with a _delete_by_query request

daniellienert:
As we have _delete_by_query in ES 5, this is a way simpler solution

#closes #333


Nikdro:
Thank you @daniellienert! 
Just tried this one for my bug (discussed in https://neos-project.slack.com/archives/C0AKFH091/p1589814741030100) 

Worked so far. No more error on change or publish. 👍

PR #341, opened 2020-07-16
BUGFIX: Fix logging of queries in case of an error

### LINKED ISSUE #340: Exceptions hidden in fetch() method
The way `logThrowable()` is used in `fetch()` makes it very hard to spot errors in queries, as no log message is ever written and an empty result is returned. One has to stumble across the exception dump by accident, more or less.

kdambekalns:
When a query fails on Elasticsearch

- it will be logged now (before being sent to Elasticsearch)
- in case of an error a system log entry will be written

Fixes #340

### DIFF
--- a/Classes/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Eel/ElasticSearchQueryBuilder.php
@@ -604,7 +604,7 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
                 $this->result['nodes'] = $this->convertHitsToNodes($searchResult->getHits());
             }
         } catch (ApiException $exception) {
-            $this->logThisQuery && $this->logger->debug(sprintf('Query Log (%s): execution failed with message: %s', $this->logMessage, $exception->getMessage()), LogEnvironment::fromMethodName(__METHOD__));
+            $this->logger->debug(sprintf('Query Log (%s): execution of query failed (see below), query was: %s', $this->logMessage, $this->request->getRequestAsJson()), LogEnvironment::fromMethodName(__METHOD__));
             $logMessage = $this->throwableStorage->logThrowable($exception);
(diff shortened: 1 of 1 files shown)


Classes/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -596,13 +598,15 @@ public function fetch(): array
...
+            $this->logThisQuery && $this->logger->debug(sprintf('Query Log (%s): execution time: %s ms -- Limit: %s -- Number of results returned: %s -- Total Results: %s', $this->logMessage, (($timeAfterwards - $timeBefore) * 1000), $this->limit, count($searchResult->getHits()), $searchResult->getTotal()), LogEnvironment::fromMethodName(__METHOD__));
 
             if (count($searchResult->getHits()) > 0) {
                 $this->result['nodes'] = $this->convertHitsToNodes($searchResult->getHits());
             }
         } catch (ApiException $exception) {
-            $this->throwableStorage->logThrowable($exception);
+            $this->logThisQuery && $this->logger->debug(sprintf('Query Log (%s): execution failed with message: %s', $this->logMessage, $exception->getMessage()), LogEnvironment::fromMethodName(__METHOD__));
```
daniellienert, line 607: Not sure, if we need two log messages here. The `logMessage` is helpful as often several queries are send when rendering a page. I would just prefix the error message with the `'Query Log (%s):` part and only send this one.

daniellienert:
Hey @kdambekalns, any thoughts on my comment?

Classes/Eel/ElasticSearchQueryBuilder.php
kdambekalns: I tweaked it a bit, but now I realize it's still not optimal.

- if `log()` was used when building the query, the query is logged before the exception log message
- without `log()` use, the exception is logged, but you don't know the query - hence I always log it now in case of an exception

This leads to exception + use of `log()` to have the query logged twice now - bad?

PR #345, opened 2020-08-10
BUGFIX: Trying to use response from exception as string

kdambekalns:
The return value of `ApiException->getResponse()` is no longer a
string, so decoding it as JSON fails.


kdambekalns:
@daniellienert I created this "on the fly", hope the branch is correct and the fix works "for all cases" (i.e. no version combinations without the PSR-7 change I think caused this…)

PR #348, opened 2020-09-04
Add indexname to log entries

DrillSergeant:
Currently the EEL query builder logs the query request body. I added the indexname as we might be in multi-index scenarios.


daniellienert (approved):
Good addition!

Issue #336, opened 2020-06-18
nodeindex:build only based on default dimension 


regniets:
We have a setup with a dimension "zz" that only shares the rootpage for language selection, but is defaultPreset.
In the latest version of the Adaptor (on Neos 5.2 & ES 7) nodeindex:build will index only the nodes that are found in the zz dimension and their respective dimension representatives. So we end up with only Home being indexed in all dimensions, but no further pages.

A change introduced that we found is in [NodeIndexCommandController](https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/0660ba2ddd14a7968436eb0d76a00b9b9b93d6c4/Classes/Command/NodeIndexCommandController.php#L388):
Instead of calling `$this->workspaceIndexer->index()`, `$this->workspaceIndexer->indexWithDimensions()` is being called, which doesn't loop over dimensions initially.

But in order for the nodes to be indexed independently in all dimensions, it still doesnt work properly, since the rootNode still is "dimensionless":
https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/0660ba2ddd14a7968436eb0d76a00b9b9b93d6c4/Classes/Indexer/WorkspaceIndexer.php#L89

daniellienert:
Uh that sounds bad. The problem was, that Dimensions were cycled twice, once in the controller and once in https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/master/Classes/Indexer/NodeIndexer.php#L308 
So I checked both loops and removed one of them. But as I understand your setup, traversing the tree is also not possible if the controller would loop over dimensions.
So how did the old version found your nodes?

regniets:
Yes, i still haven't found out why and how it worked in the old version.

The new approach seems to work only if you have 1:1 relations in dimensions and nodes, but doesn't take "loose dimensions" into account. I can provide you with a PR how we worked around this issue (which is very pragmatic, not sure if it's the right way to go) and might make the second loop redundant (you might lose some weird orphan nodes though).

daniellienert:
[...]

PR #349, opened 2020-09-04
Decouple tests from nodetype-package, add test for multi-dimension-scenario

DrillSergeant:
- added nodetype definitions for a document and a content nodetype
- use those nodetypes in existing tests (instead on reliying on the existence of the nodetype-package)
- add indexer test for multi-dimension scenario (fails currently as excpected)


DrillSergeant:
This is related to #336. It adds a test that will fail to be ablte to reproduce the existing issue.

daniellienert:
Features of this PR were already included in #350

PR #371, opened 2021-02-05
TASK: Add mapping for Elasticsearch 7.x to settings

### LINKED ISSUE #366: Elasticsearch 7.x mappings missing
Version 7 and 8 of this package support Elasticssearch 7, so I would expect the following configuration to work:
```
Flowpack:
  ElasticSearch:
    ContentRepositoryAdaptor:
      driver:
        version: 7.x
```

It seems that Elasticsearch 6 and 7 mapping is the same, so a mapping for 7.x should be added here:
https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/master/Configuration/Settings.yaml#L15

kdambekalns:
Fixes #366

### DIFF
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -14,7 +14,7 @@ Flowpack:
       driver:
         version: '6.x'
         mapping:
-          6.x:
+          6.x: &v6x
             query:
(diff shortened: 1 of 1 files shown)


kdambekalns:
The build failure seems unrelated, I tested this locally with success…

daniellienert (approved):

PR #373, opened 2021-02-11
BUGFIX: Set up missing alias during node indexing

### LINKED ISSUE #370: Indexing creates an index instead of the expected alias
When running `nodeindex:build` and no indexes exist at all, an index named `foo-postfix` is created. But instead of an alias `foo` pointing to that, an actual index `foo` is created and filled with data.

This leads to at least two problems:

- when aliases are about to be switched the next time, an error occurs, since the alias (that isn't an alias) cannot be deleted
- mapping is not as expected but built dynamically (see #356)

kdambekalns:
Fixes #370


kdambekalns:
Not sure about lower branches… I am quite certain I saw this issue on older versions, but don't have a project at hand right now to verify. Also older versions (of Elasticsearch) are EOL anyway.

daniellienert (approved):
Looks good by reading, thanks Karsten

kdambekalns:
@daniellienert – You mentioned fixed tests on Slack, should that help with this PR? Any hints on what needs to be fixed?

daniellienert:
@kdambekalns rebase it in 7.0 should fix the Tests

kdambekalns:
> rebase it in 7.0 should fix the Tests

That was easy… now I just needed to remove `JSON_THROW_ON_ERROR` again for PHP 7.2. 🤯

Issue #375, opened 2021-03-04
Explosion of mapping fields


DrillSergeant:
Since version 7.0.5 elasticsearch field mapping for neos_fulltext_parts gets flooded:

```
                "neos_fulltext_parts": {
                    "properties": {
                        "01b4f10b-c7f8-4e37-9c0a-2e04bca0df54": {
                            "properties": {
                                "h1": {
                                    "type": "text",
                                    "fields": {
                                        "keyword": {
                                            "type": "keyword",
                                            "ignore_above": 256
                                        }
                                    }
                                }
                            }
                        },
                        "02d557fc-887b-4d79-bde5-5d5db6f0eaad": {
                            "properties": {
                                "h2": {
                                    "type": "text",
                                    "fields": {
                                        "keyword": {
                                            "type": "keyword",
                                            "ignore_above": 256
                                        }
                                    }
                                },
                 [this repeats for every existing node as far as I can see, lots of entries here]
```

This leads to this error during indexing:
 ```
 Error:
=======

{
    "update": {
        "_index": "neoscr-fb11fdde869d0a8fcfe00a2fd35c031d-1614847774",
        "_type": "_doc",
        "_id": "3eddf8a0b104c5169e09a18a040916e27d60078a",
        "status": 400,
        "error": {
            "type": "illegal_argument_exception",
            "reason": "Limit of total fields [1000] in index [neoscr-fb11fdde869d0a8fcfe00a2fd35c031d-1614847774] has been exceeded"
        }
    }
```

Reason for this is: the introduced array_filter here: https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/compare/7.0.4...7.0.5#diff-433af2c73384cf274bd3eb5a72387db03cc0816ba92b69e59d14d1bd71464592R76

That filters all array-values that are boolean false out of the mapping-config, but for the field "neos_fulltext_parts" the config "enabled: false" is configured and I think it should not be filtered out.

```
'Neos.Neos:Document':
  search:
    fulltext:
      isRoot: true
  properties:
[...]
      search:
        elasticSearchMapping:
          type: object
          enabled: false
        indexing: ''
 ```

kdambekalns:
[...]

PR #372, opened 2021-02-05
TASK: Allow to unset elasticSearchMapping parameters

kdambekalns:
Replicates 97aa56fe2b1f30e16b53c766c2ecda954a5cae08 for the v7 branch.

PR #376, opened 2021-03-05
TASK: Adjust filter allowing to unset elasticSearchMapping parameters

kdambekalns:
A follow-up to #372: Only filter null values, but keep false and any
other value considered `empty()` by PHP.

Related #375


DrillSergeant:
Thank you for that. I just tested it and can confirm it fixes #375. Would be nice to have a new release that I can rollout to our fleet.

daniellienert (approved):

Issue #353, opened 2020-09-10
Moving a node leads to double-indexed nodes


daniellienert:
Currently, the identifier of the elasticsearch document is calculated using the nodepath.
When a node is moved, a new document with a new identifier is added to the index instead of updating the existing.

baschny:
why not use the identifier instead of the node-path when finding the nodes to delete/create on move? the identifier doesn't change when moving it, right?

dfeyer:
> why not use the identifier instead of the node-path when finding the nodes to delete/create on move? the identifier doesn't change when moving it, right?

Don't remember why we use the context path for this document identifier ... sounds odd ... but some memories could be nice .... maybe in the past the node uuid change during a move operation ?

baschny:
We've been running a "patched" version of this that does exactly this (use the Identifier) without a problem. I thought we did a pull request back then, but apparently not. Maybe @remuslazar can provide a pull request?

remuslazar:
@baschny sure thing!

remuslazar:
@baschny @dfeyer done, see #378

PR #378, opened 2021-04-10
Use Node Identifier, Fix ES Duplicates 

remuslazar:
## Abstract

This changes the way the Elasticsearch document identifier is being calculated by using the node identifier instead of the node path.

## Resolves

* https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/issues/353

## Rationale

Using the node path to deduce the identifier in Elasticsearch has some flaws and bad side effects when moving nodes around; in that case, especially when using an async node indexing (e.g. by using the [ES Queue Indexer](https://www.neos.io/de/download/pakete/flowpack/flowpack-elasticsearch-contentrepositoryqueueindexer.html)), the indexing process will produce duplicates, basically it will leave the old node around and index a new node (with the correct path) and this will cause race conditions when doing ES queries and also issues when doing e.g. pagination etc.

## Breaking

This change will **require** a "flow:nodeindex", so maybe a minor (or major?) release will the way to go. 

## Implementation Details

In addition to the logic change this also makes the existing implementation more DRY by removing the duplicated code in the `IndexerDriver` - see 1baf54e.

## Testing

This also adds an implementation for the existing `NodeIndexerTest::nodeMoveIsHandledCorrectly()` test method.

To run the functional tests in e.g. docker, do:

```shell
FLOW_CONTEXT="Testing/ElasticVersion6" bin/phpunit --colors --stop-on-failure \
  -c Build/BuildEssentials/PhpUnit/FunctionalTests.xml \
  Packages/Application/Flowpack.ElasticSearch.ContentRepositoryAdaptor/Tests/Functional
```

## Discussion

I will like to port this change to older versions as well so this will also work in older Neos 4.3 projects..

### DIFF
--- a/Classes/Driver/Version6/IndexerDriver.php
+++ b/Classes/Driver/Version6/IndexerDriver.php
@@ -13,7 +13,7 @@ namespace Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\Version6;
  * source code.
  */
 
-use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Domain\Model\TargetContextPath;
+use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Indexer\NodeIndexer;
 use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\AbstractIndexerDriver;
 use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\IndexerDriverInterface;
 use Flowpack\ElasticSearch\Domain\Model\Document as ElasticSearchDocument;
@@ -90,11 +90,7 @@ class IndexerDriver extends AbstractIndexerDriver implements IndexerDriverInterf
             return [];
         }
 
-        $closestFulltextNodeContextPath = $closestFulltextNode->getContextPath();
-        if ($targetWorkspaceName !== null) {
-            $closestFulltextNodeContextPath = (string)(new TargetContextPath($node, $targetWorkspaceName, $closestFulltextNodeContextPath));
-        }
-        $closestFulltextNodeDocumentIdentifier = sha1($closestFulltextNodeContextPath);
+        $closestFulltextNodeDocumentIdentifier = NodeIndexer::calculateDocumentIdentifier($closestFulltextNode);
 
         if ($closestFulltextNode->isRemoved()) {
             // fulltext root is removed, abort silently...
--- a/Classes/Indexer/NodeIndexer.php
+++ b/Classes/Indexer/NodeIndexer.php
@@ -243,7 +243,7 @@ class NodeIndexer extends AbstractNodeIndexer implements BulkNodeIndexerInterfac
                 }
             }
 
-            $documentIdentifier = $this->calculateDocumentIdentifier($node, $targetWorkspaceName);
+            $documentIdentifier = NodeIndexer::calculateDocumentIdentifier($node, $targetWorkspaceName);
             $nodeType = $node->getNodeType();
 
             $mappingType = $this->getIndex()->findType($nodeType->getName());
@@ -346,11 +346,11 @@ class NodeIndexer extends AbstractNodeIndexer implements BulkNodeIndexerInterfac
      * Returns a stable identifier for the Elasticsearch document representing the node
      *
      * @param NodeInterface $node
-     * @param string $targetWorkspaceName
+     * @param null $targetWorkspaceName
      * @return string
      * @throws IllegalObjectTypeException
      */
-    protected function calculateDocumentIdentifier(NodeInterface $node, $targetWorkspaceName = null): string
+    public static function calculateDocumentIdentifier(NodeInterface $node, $targetWorkspaceName = null): string
     {
         $workspaceName = $targetWorkspaceName ?: $node->getWorkspace()->getName();
         $nodeIdentifier = $node->getIdentifier();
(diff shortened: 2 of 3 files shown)


baschny:
👍 this has been working for us in a big site for the last 5 years successfully.

baschny:
As a side-note: we (@cron-eu) will be using this in "5.0" branch for our use-case, so we could provide a PR with the fix even for that low version.

An idea to have this included would be to release it as "5.1.0" (and 6.1.0, 7.1.0, 8.1.0) and making a note that a "reindex" is required for that minor update. Don't know if one can expect people to read docs on that for a "simple patch level" release. And I don't think there is a "backwards compatible way" of fixing the bug without reindexing your content.

dfeyer (approved):
By reading the code, good job and nice move. Now we need to take a decision on the major or minor release strategy ;)

Classes/Indexer/NodeIndexer.php
```diff
@@ -274,6 +274,13 @@ public function indexNode(NodeInterface $node, $targetWorkspaceName = null): voi
         $handleNode = function (NodeInterface $node, Context $context) use ($targetWorkspaceName, $indexer) {
             $nodeFromContext = $context->getNodeByIdentifier($node->getIdentifier());
             if ($nodeFromContext instanceof NodeInterface) {
+                if ($node->getPath() !== $nodeFromContext->getPath()) {
+                    // If the node from context does have a different path, purge the context cache and re-fetch
```
kdambekalns, line 278: A note explaining when/why that could happen would be awesome!

Classes/Indexer/NodeIndexer.php
```diff
@@ -274,6 +274,13 @@ public function indexNode(NodeInterface $node, $targetWorkspaceName = null): voi
...
+
+                    // TODO: find a better way to handle this
```
kdambekalns, line 280: What is wrong with this solution? Would you like to avoid flushing the full cache?

kdambekalns (commented):
Love this, left some comments about more explanation of things.

The release could be minor IMHO, and yes, I at least expect people to read things. 😎

And backporting that to older branches would be great, IMHO! 🚀

Classes/Indexer/NodeIndexer.php
remuslazar: Indeed; I will prefer flushing the cache just for this specific node, but I was not able to figure out a good way to handle that. Do you have an idea here?

And maybe it should be safe(r) to just flush the cache **always**, regardless of the "node path was changed" condition. What do you think?

### TL;DR

I can remember that back then, in 2016 (!) as I have implemented this initially, it took me some time of hair pulling to track down this issue.. I still do not quite understand why the cached data does not cause any issues when the node path was not being changed; maybe the neos context node caching behavior has some flaws, using the node path for invalidation or stuff like that?

Classes/Indexer/NodeIndexer.php
remuslazar: see here my next comment on that topic.

dfeyer:
Any news on this one ?

kdambekalns:
I just had a quick look at the chanegs again and still like it. Nothing new, though. ;)

kdambekalns:
Add @kitsunet as reviewer, lot's of ES usage "in his world", so maybe he can check.

Classes/Indexer/NodeIndexer.php
```diff
@@ -339,19 +346,16 @@ protected function toBulkRequest(NodeInterface $node, array $requests = null): v
...
-        $contextPath = $node->getContextPath();
-
-        if ($targetWorkspaceName !== null) {
-            $contextPath = (string)(new TargetContextPath($node, $targetWorkspaceName, $contextPath));
-        }
+        $workspaceName = $targetWorkspaceName ?: $node->getWorkspace()->getName();
+        $nodeIdentifier = $node->getIdentifier();
 
-        return sha1($contextPath);
+        return sha1($nodeIdentifier . $workspaceName);
```
daniellienert, line 358: Currently, the contextPath contains, site, workspace, dimensions, and nodeIdentifier.
Dimensions should/may be covered, as the are going int separate indices. But sites are currently combined in the indices by dimension. Could that be an issue when cloning a site where somehow the nodeIdentfiers are the same. 

Why don't we just use the persistent_object_identifier, which is already unique?

Classes/Driver/Version6/IndexerDriver.php
```diff
@@ -13,7 +13,7 @@
  * source code.
  */
 
-use Flowpack\ElasticSearch\ContentRepositoryAdaptor\Domain\Model\TargetContextPath;
```
daniellienert, line 16: This class should be obsolete now, and should be removed.

daniellienert (changes requested):

Classes/Indexer/NodeIndexer.php
kdambekalns: If you have two sites with nodes sharing the same identifier, you are doomed, I think. That should never happen, no? Then again, it could happen, so adding the site here might be good.

Using the `persistent_object_identifier` OTOH is a no-go IMHO. It should stay an implementation detail internal to the CR and not be used outide it.

dfeyer:
@remuslazar Any news on this PR ? Should be nice to have it merged

daniellienert:
Dear @remuslazar, 

thanks for tackling this in the first place! During the core sprint, I reviewed and tested the change and it worked perfectly. We decided though, that we like to release the change as a minor version and so I implemented the feature as an alternative strategy in https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/394

I already merged that feature. And will close this one in favor of #394

baschny:
@daniellienert great stuff, thanks a lot for your effort! 👍

Issue #380, opened 2021-04-13
Hard coded ContextFactory


bwaidelich:
This package uses the `ContextFactory` of the ContentRepository package in several places where it should use the `ContextFactoryInterface` instead.

That leads to failures like
```
Call to undefined method Neos\ContentRepository\Domain\Service\Context::getCurrentSiteNode() [...]
```
when expecting a *Neos*  context during indexing.

PR #381, opened 2021-04-13
BUGFIX: Replace hard coded ContextFactory references with the interface

bwaidelich:
This replaces the use of the `ContextFactory` of the ContentRepository package in several places with the `ContextFactoryInterface` instead.

That fixes failures like
```
Call to undefined method Neos\ContentRepository\Domain\Service\Context::getCurrentSiteNode() [...]
```
when expecting a *Neos*  context during indexing.

Fixes: #380


kdambekalns (approved):
Looks good, thanks for finding and fixing!

daniellienert (approved):

PR #383, opened 2021-09-28
TASK: Configure ignore_above for keyword fields

daniellienert:
This sets ignore_above of keyword fields to 8191 according to
https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html#ignore-above
which avoids "Document contains at least one immense term in field" errors
when indexing fields containing lots of text.


kdambekalns (approved):
Probably does no harm, even though this might mask a valid concern with indexing data…

daniellienert:
I feel you. But you never know what users put into text fields and "keywords" with > 8k characters are probably not what you wanted.

PR #384, opened 2021-09-28
TASK: Make the size of ingested attachment data configurable

daniellienert:
This also fixes a problem with huge files exceeding available ram.

### DIFF
--- a/Classes/AssetExtraction/IngestAttachmentAssetExtractor.php
+++ b/Classes/AssetExtraction/IngestAttachmentAssetExtractor.php
@@ -47,7 +47,7 @@ class IngestAttachmentAssetExtractor implements AssetExtractorInterface
     protected $logger;
 
     /**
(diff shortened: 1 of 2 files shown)


Classes/AssetExtraction/IngestAttachmentAssetExtractor.php
```diff
@@ -126,7 +132,7 @@ protected function getAssetContent(AssetInterface $asset): string
         }
 
         stream_filter_append($stream, 'convert.base64-encode');
-        $result = stream_get_contents($stream);
+        $result = stream_get_contents($stream, (int)$this->indexedChars);
```
kdambekalns, line 135: Does that work reliably? Binary content, just split at an arbitrary position is indexed correctly? I'd have thought that breaks…

daniellienert: Good point. Currently testing in production. 
Alternatively there should be an option to exclude files > <givenFileSize> entirely

daniellienert: Yep. Is borken.

PR #385, opened 2021-09-28
FEATURE: Define max file size for ingesting attachments

daniellienert:
Huge files may exceed the avalaible memory and should thus
not be ingested.


daniellienert:
@kdambekalns, this approach actually works, wdyt

kdambekalns (approved):
That looks good (by reaading) - thanks!

Issue #367, opened 2021-01-28
Migration 20200513223401 creates wrong result


kdambekalns:
After running this migration, I see `_hiddenInIndex` being replaced by `neosneos_hidden_in_index`:

<img width="246" alt="Screenshot 2021-01-28 at 20 15 41" src="https://user-images.githubusercontent.com/95873/106187261-a9ba7780-61a5-11eb-88ee-6bcca41e661a.png">

It should result in `neos_hidden_in_index`, though.

Update: And `__parentPath` becomes `neos_parent_path ` (with a trailing space!)

kitsunet:
This whole migration is way to aggressive. it just turend a customer instance into totally broken, because also a `.children([_hidden = true])` was changed as well has a something_something_hidden in their own package sql migrations and ORM index annotations. I would actually say we should remove this migration as the _useful_ case is definitely rarer IMHO than the "wrong" case.

daniellienert:
Hey guys, sorry for that. Thats really bad. Lets remove it. I'd supply a patch.

PR #387, opened 2021-12-01
TASK: Remove migration for renaming underscore prefixed fields

daniellienert:
The migration is too aggressive and also renames
valid flow queries.

Fixes: #367


kdambekalns (approved):
Burn it with fire. 😜

What would be great, though: Provide that list of name pairs for manual use! If it's already in some upgrade instructions, I take the blame for bot checking. 😎

kitsunet (approved):
Thanks for taking care <3

PR #388, opened 2022-01-07
BUGFIX: Keep fulltext index of hidden nodes as those may be searched in backend use cases

mficzel:
The fulltext fields for hidden nodes previously were calculated but then not stored. This makes sense for removed nodes but hidden nodes exist and may be found by users that have the correct permissions.

### DIFF
--- a/Classes/Driver/Version6/IndexerDriver.php
+++ b/Classes/Driver/Version6/IndexerDriver.php
@@ -126,7 +126,7 @@ class IndexerDriver extends AbstractIndexerDriver implements IndexerDriverInterf
                             ctx._source.neos_fulltext_parts = new HashMap();
                         }
 
-                        if (params.nodeIsRemoved || params.nodeIsHidden || params.fulltext.size() == 0) {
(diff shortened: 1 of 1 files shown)


mficzel:
This will probably need a rebase when accepted as lower branches are affected aswell.

dfeyer (approved):
Make sense

Classes/Driver/Version6/IndexerDriver.php
```diff
@@ -126,7 +126,7 @@ public function fulltext(NodeInterface $node, array $fulltextIndexOfNode, string
                             ctx._source.neos_fulltext_parts = new HashMap();
                         }
 
-                        if (params.nodeIsRemoved || params.nodeIsHidden || params.fulltext.size() == 0) {
```
daniellienert, line 129: If content from hidden nodes are added to the fulltext_parts, I assume they are automatically merged to the parent document where it is impossible to distinguish between content from hidden and not hidden node and the document would also be found by the originally hidden content.

PR #392, opened 2022-04-13
BUGFIX: Include all base workspaces in neos_workspace

### LINKED ISSUE #391: Querybuilder misses non-live base workspaces
### What is broken?
If you are using an custom workspace (not live) as your users base-workspace, no nodes from that workspace are returned. 

### Why?
If you look at the ES-Query generated by a user in Backend with non-live base-workspace, you'll see the following condition: 
```json
{
    "terms": {
        "neos_workspace": [
            "live",
            "user-ndroste"
        ]
    }
},
```
This happens, because the builder only adds the current workspace `user-ndroste` and the `live` (hard-coded) workspace to the query. See Code: ` array_unique(['live', $contextNode->getContext()->getWorkspace()->getName()])]);` at https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/blob/master/Classes/Eel/ElasticSearchQueryBuilder.php#L830

### Fix?! 
We can just add all BaseWorkspaces from the $contextNode: 
```php
        $workspaces = array_merge(
            [$contextNode->getContext()->getWorkspace()->getName()],
            array_keys($contextNode->getContext()->getWorkspace()->getBaseWorkspaces())
        );
```

This leads to the following terms in query: 
```json
{
    "terms": {
        "neos_workspace": [
            "user-ndroste",
            "tools-zp8z8",
            "live"
        ]
    }
},
```
I can't see any disadvantages with the order of the workspaces here. But if there are some, we could easily turn the order generated here around.

PR with fix is on it's way.

Nikdro:
Fixes #391

### DIFF
--- a/Classes/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Eel/ElasticSearchQueryBuilder.php
@@ -825,9 +825,13 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
             ]
         ]);
 
+        $workspaces = array_merge(
+            [$contextNode->getContext()->getWorkspace()->getName()],
+            array_keys($contextNode->getContext()->getWorkspace()->getBaseWorkspaces())
+        );
         //
         // http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-terms-filter.html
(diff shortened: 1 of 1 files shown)


dfeyer (approved):
Thanks for the contribution

daniellienert (approved):
Makes sense to me

PR #394, opened 2022-05-04
BUGFIX: Moving nodes must not lead to duplicated documents 

daniellienert:
Original PR by [remuslazar](https://github.com/remuslazar).

## Abstract

This changes the way the Elasticsearch document identifier is being calculated by using the node identifier instead of the node path.

## Resolves

* https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/issues/353

## Rationale

Using the node path to deduce the identifier in Elasticsearch has some flaws and bad side effects when moving nodes around; in that case, especially when using an async node indexing (e.g. by using the [ES Queue Indexer](https://www.neos.io/de/download/pakete/flowpack/flowpack-elasticsearch-contentrepositoryqueueindexer.html)), the indexing process will produce duplicates, basically it will leave the old node around and index a new node (with the correct path) and this will cause race conditions when doing ES queries and also issues when doing e.g. pagination etc.

## Breaking

This change will **require** a "flow:nodeindex", so maybe a minor (or major?) release will the way to go. 

## Implementation Details

In addition to the logic change this also makes the existing implementation more DRY by removing the duplicated code in the `IndexerDriver` - see 1baf54e.

## Testing

This also adds an implementation for the existing `NodeIndexerTest::nodeMoveIsHandledCorrectly()` test method.

To run the functional tests in e.g. docker, do:

```shell
FLOW_CONTEXT="Testing/ElasticVersion6" bin/phpunit --colors --stop-on-failure \
  -c Build/BuildEssentials/PhpUnit/FunctionalTests.xml \
  Packages/Application/Flowpack.ElasticSearch.ContentRepositoryAdaptor/Tests/Functional
```


PRGfx:
When I change the DocumentIndentifierGenerator as described, all nodes seem to be indexed with the last workspace handled in `nodeindex:build`. When I query the index in the end, I only get nodes for the last workspace in the list printed by that command.
Querying without a filter on workspace only returns these nodes (with the default identifier generator I get the nodes for each workspace respectively).
`nodeindex:build --workspace ...` populates the index for that workspace, but as expected every other workspace is missing.

Am I missing something?

PR #395, opened 2022-06-17
FEATURE: Support `.simpleQueryStringFulltext()` simple query string query for fulltext search

puchsteink:
We want to support phrase searches like `"Neos and Flow"`, where
quotes mean "I want this exact phrase" (similar to many search engines).

When using the .fulltext() query, this did not work, and we sometimes
encountered crashes when the query string contained quotes, f.e. `"Hallo"`
or `Ha"llo"`.

This is because the .fulltext() query does some magic for supporting UTF8
(as it seems), namely: `trim(json_encode($searchWord, JSON_UNESCAPED_UNICODE), '"')`.
This removes some quotes, but not all of them :-)

The simple_query_string query of Elasticsearch exposes a query string syntax
which the user is directly allowed to enter:

> While its syntax is more limited than the query_string query, the simple_query_string
> query does not return errors for invalid syntax. Instead, it ignores any invalid
> parts of the query string.

This is exactly what we want for a good search field behavior.

NOTE: We created an additional method because we were unsure about breaking
      backwards compatibility; but we think `simpleQueryStringFulltext()` should
      be used almost universally as replacement for `fulltext()`, as it returns
      better results and does not throw exceptions..


kdambekalns (approved):
Looks good… I wonder how this developed in the past. Is `simple_query_string` something "new", did it not exist when the `fulltext(…)` methods was invented?

daniellienert (commented):
No objections to add that. It also was previously asked for. Please add the a line of documentation to the readme. 

Just read through the features of both methods and they seem to have nearly the same capabilities. simpleQueryString seems to lack the "fuzziness" option, which I use in projects while having the other fuzzy_* options. Maybe its just missing in the docs. (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html)

For reference. In our projects we use https://github.com/Flowpack/Flowpack.SearchPlugin/blob/master/Classes/Utility/SearchTerm.php#L26 for safe sanitizing and configure the `default_operator` to AND and are quite happy with the results.

skurfuerst:
@kdambekalns I probably didn't know better when initially inventing fulltext() :-)

skurfuerst (approved):

PR #398, opened 2022-09-28
FEATURE: Add Node-Type Filter

erkenes:
Create a method which allows to create a nodetype filter to include and exclude specific nodetypes


erkenes:
Simplified example:

```php

public function getVehiclePages(NodeInterface $siteNode)
{
        $filter = [
                'Foo.Bar:Document.Vehicle',
                'Foo.Bar:Document.Motorcyle',
                '!Neos.Neos:Document' // this nodetype will be excluded
        ];

        $queryBuilder = new ElasticSearchQueryBuilder();
        $queryBuilder->query($siteNode);
        $queryBuilder->nodeTypeFilter(implode(',', $filter));
        
        // [...]
}
```

erkenes:
@Sebobo @daniellienert can you review this PR?

Classes/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -131,6 +138,97 @@ public function nodeType(string $nodeType): QueryBuilderInterface
...
 
+    /**
+     * Filter multiple node types
+     *
+     * @param string|array $nodeTypeFilter
+     * @return ElasticSearchQueryBuilder
+     * @throws QueryBuildingException
+     * @api
+     */
+    public function nodeTypeFilter($nodeTypeFilter): QueryBuilderInterface
```
kitsunet, line 149: These days I would rather not implement such a "messy" interface anymore, but instead have a signature like
`nodeTypeFilter(array $expectedNodeTypes, array $excludedNodetypes)` (or introduce a filter DTO but that then would probably have to reside in the CR rather than here).

erkenes: You're right two parameters are cleaner :)
I updated the PR.

Classes/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -131,6 +131,55 @@ public function nodeType(string $nodeType): QueryBuilderInterface
...
+     *
+     * @param array $expectedNodeTypes NodeTypes that should be expected
+     * @param array $excludedNodeTypes NodeTypes that should be excluded
+     * @return ElasticSearchQueryBuilder
+     * @throws QueryBuildingException
+     * @api
+     */
+    public function nodeTypeFilter(array $expectedNodeTypes, array $excludedNodeTypes): QueryBuilderInterface
+    {
+        $excludeShould = [];
```
daniellienert, line 145: Nitpick: It ends in an bool must / must_not query so should is a bit missleading here

Classes/Eel/ElasticSearchQueryBuilder.php
```diff
@@ -131,6 +131,55 @@ public function nodeType(string $nodeType): QueryBuilderInterface
...
+    /**
+     * Filter multiple node types
+     *
+     * @param array $expectedNodeTypes NodeTypes that should be expected
+     * @param array $excludedNodeTypes NodeTypes that should be excluded
+     * @return ElasticSearchQueryBuilder
+     * @throws QueryBuildingException
+     * @api
+     */
+    public function nodeTypeFilter(array $expectedNodeTypes, array $excludedNodeTypes): QueryBuilderInterface
```
daniellienert, line 143: Would be nice if you would add a slim test in `ElasticSearchQueryTest`.

daniellienert (commented):
Hey @erkenes, thanks a lot for the improvement! 
I have two comments tho.
Cheers Daniel

Classes/Eel/ElasticSearchQueryBuilder.php
erkenes: I added a test and adjusted the existing tests to match the new circumstances.

Classes/Eel/ElasticSearchQueryBuilder.php
erkenes: I changed `Should` to `Terms`

PR #400, opened 2022-12-06
BUGFIX: ElasticSearchQueryBuilder removes json encode generated quotes

gjwnc:
By default the ES CR Adaptor uses [query_string](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-syntax) which allows some search syntax in query string.

In the current implementation, `json_encode` is used on the query string which adds double quotes around the search term,
e.g. `brown fox` is json encoded to `"brown fox"`. Then `trim` is used to remove the `"`.

Since one can use double quotes in the search term itself, e.g. `"brown fox"` this results in json encoded `"\"brown fox\""`. If you use `trim` you get invalid json which causes an exception because of invalid syntax in the ES query: `\"brown fox\`.

This PR, just removes the outer quotes created by `json_encode`.

### DIFF
--- a/Classes/Eel/ElasticSearchQueryBuilder.php
+++ b/Classes/Eel/ElasticSearchQueryBuilder.php
@@ -684,8 +684,11 @@ class ElasticSearchQueryBuilder implements QueryBuilderInterface, ProtectedConte
      */
     public function fulltext(string $searchWord, array $options = []): QueryBuilderInterface
     {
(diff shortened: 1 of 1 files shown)


gjwnc:
I just noticed, that this issues is mentioned here #311 . Since, it could depend on the use case, both PR's might be considered independent.

PR #402, opened 2023-02-13
Bugfix: Should throw error if indexname is not set

amtee:
It is common practice to use an ENV variable to set the index name. If it's not present,It is common practice to use an ENV variable to set the index name. If it's not present, `$name` is not empty string, but `false`. 

I know this is likely not a bug in the Adaptor itself, maybe let's see this as an enhancement 😄  $name is not empty string, but false.


daniellienert (changes requested):
Hey @amtee, 
According to https://www.php.net/manual/de/function.trim.php, trim never returns `false` - so your check will never be true. Do I miss somehing?

PR #406, opened 2023-08-29
Draft: Feature elasticsearch 8 support

marcrobertscamao:
As Elasticsearch 8 has removed types the _type property must be removed from the indexing. Everything else appears to have stayed the same. I have not yet had time to test this in detail, however it appears to work with a simple instance we are currently building. I am opening this so that hopefully noone needs to duplicate work.

Unfortunately for this to work we also need to adapt the Flowpack.Elasticsearch/Domain/Model/Mapping to remove the 'include_type_name' option. This is a much larger issue however.

### DIFF
--- a/Classes/Driver/Version8/DocumentDriver.php
+++ b/Classes/Driver/Version8/DocumentDriver.php
@@ -1,0 +1,45 @@
+<?php
+declare(strict_types=1);
+
+namespace Flowpack\ElasticSearch\ContentRepositoryAdaptor\Driver\Version8;
+
+/*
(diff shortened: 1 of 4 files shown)


daniellienert (changes requested):
Hey @marcrobertscamao, 

Thank you very much for your work! The change works and is leaving the current behaviour untouched. 

I am sure, that the "_type" argument was just a leftover from much older versions and is not needed, at leas in ES6 upwords. So I simplified everything by just removing the type value in https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/pull/407 and fixed the tests to run green for all versions.

Issue #377, opened 2021-03-19
Search result shows hidden content parts


mariofischer:
We updated flowpack/elasticsearch-contentrepositoryadaptor to 7.0.6 (from 6.0.6). In the search result there are now content-parts (elements from the content tree) shown which are hidden in Neos (Before that the result did not show the hidden parts).
These elements are part of the indexed fulltext.

mariofischer:
Additional clarification: The problem seems to be that the content of a node is being indexed into the page fulltext even if the parent node is hidden.

dfeyer:
This issue duplicate this one:
https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/issues/214#issuecomment-810469614

kdambekalns:
Am no longer sure this is a duplicate. Even though #214 is quite old, I am seeing this issue now after an upgrade of ESCR from 6 to 7 – now hidden nodes are returned that were not found before.

Fact: Both indexes (old and new) do contain data about the nodes, and the data itself does not differ (both are not marked as hidden in the index.)

The question thus is: Did the code fetching the nodes from the ES result change in that regard? It does not look like it for me: https://github.com/Flowpack/Flowpack.ElasticSearch.ContentRepositoryAdaptor/compare/6.0.6...7.0.9#diff-5310f1e603b84f1f7463df8670be1896e0cf5f7c3a32b4aa0a8f2a374c5eb6c4L833-L872

Does that mean the CR changed it's behaviour? But I did not update that… 🤔

daniellienert:
Hey @kdambekalns, thats quite interesting. Obviously we have several related problems here: 
- Hidden nodes, found by ES are resolved by the CR (The hack, why it kind of worked before)
- Nodes, hidden by there parent are found in general 
- Fulltext extracts of nodes, hidden by there parent is indexed and even worse, displayed in result snippets

I am also pretty sure, that this can be solved with current CR as somehow flowquery is also capable to handle it correctly.

Wat do you think about a brainstorming session to discuss some approaches?

kitsunet:
I would also be interested then :)

kdambekalns:
> Wat do you think about a brainstorming session to discuss some approaches?

It would be a pleasure. Suggest a date/time starting this Friday that suits you. :)

TheLalaMan:
Any news on this? Our editors also noticed this recently...

[...]

Issue #214, opened 2017-06-02
Fulltext indexing for nodes "hidden through parent" is not correct


kdambekalns:
Given:

- a node that is contained in a container, e.g. a multi-column element, containing text "Foobar"
- that container is marked as hidden
- the fulltext index is built

Problem:

If the document that contains this node is found using the FLowpack.SearchPlugin, the content excerpt contains the "Foobar" string, even though it's not visible in the rendered page

Conclusion:

The indexer checks each node for it's hidden state, but since the node containing "Foobar" is _not hidden by itself_, but only invisible because it's container is hidden, it is indexed and the extracted fulltext parts are added to the fulltext root (document node).

kdambekalns:
Now, fixing this seems hard. It would mean we need to check the nodes on the path to the site node for every node we index. Given the performance issues with indexing we have already, this seems not to be a viable solution.

Any other ideas or clever solutions highly welcome!

dfeyer:
That's something we can not solve before the new CR unfortunatly ...

kdambekalns:
Not sure how the new CR would solve that!?

PRGfx:
For updating the complete index I tried using an aspect memoizing node visibility per context-path as a workaround for a small project. That did work quite nicely however ES is probably mostly used for larger projects...

I think there is a second notable problem in updating the index for a sub-tree e.g. when hiding a document with real-time indexing. Especially toggling visibility for a large sub-tree would be quite intense I imagine. Adding and removing the nodes for every change doesn't seem viable. Maybe one could work with some `in_visible_path` property\*. (Using something like an SQL join with a `(contextpath, visible)` doesn't seem to be the elasticsearch way.)

I currently see these three scenarios:
* real-time indexing; a node is set to invisible:
  this state could be applied to the whole sub-tree (would something like that be possible? update where path like ...%?)
* real-time indexing; a node is set to visible:
  the sub-tree needs to be traversed from the "top" down to the first invisible node to reset `in_visible_path`
* building the index
  the nodes are indexed "from the bottom" (or anywhere in between) and the parents would need to be traversed (at least once)


Probably this whole thing should be tied to some other configurable property of a node, as one maybe wants to exclude an otherwise visible node from search-results (i.e. some "internal" part of the website) without setting a checkbox on each sub-page, while still indexing the nodes to maybe populate list elements?

[...]

PR #413, opened 2025-02-10
BUGFIX: Skip content with hidden parent nodes in fulltext index

Sebobo:
With this change nodes with hidden parents of type content are skipped during indexing. 
By checking the visible state, the hidden flag and the time based visibility are checked.
Document nodes, which are full text roots are already checked, so we don't need to traverse the tree to the site node.
The performance impact should be negligible as the parent nodes are usually already in the cache.

As the fulltext indexing is a snapshot to make the time based visibility fully work. There needs to be a scheduled reindexing.

Relates: #214 #377

### DIFF
--- a/Classes/Indexer/NodeIndexer.php
+++ b/Classes/Indexer/NodeIndexer.php
@@ -243,6 +243,16 @@ class NodeIndexer extends AbstractNodeIndexer implements BulkNodeIndexerInterfac
                 }
             }
 
+            // Don't index node into fulltext if it has a hidden parent node that is not a fulltext root
+            // Example: If a text node is in a hidden container element, it should not be indexed into the document
+            $parentNode = $node;
+            while (($parentNode = $parentNode->getParent())
+                && !$parentNode->getNodeType()->getConfiguration('search.fulltext.isRoot')) {
+                if (!$parentNode->isVisible()) {
+                    return;
+                }
+            }
+
             $documentIdentifier = $this->documentIdentifierGenerator->generate($node, $targetWorkspaceName);
             $nodeType = $node->getNodeType();
 


dlubitz:
This would change the current behavior, as currently also hidden nodes get indexed - and filtererd out on quering. 

With this change we would have a mix of nodes which might be hidden itself, but no nodes with hidden parents. Also I'm not sure how this would work with re-indexing on change of a given node. You would need to update all child nodes too, to get this working without recreating this index.

From my POV there is no easy way to fix this issue, because there is a problem in the concept.

Sebobo:
Fulltext indexing always includes all nodes. To prevent side effects from this change we could only apply the filter to nodes that are not marked as root in their search config.

dlubitz:
> Fulltext indexing always includes all childnodes

Content nodes yes, but not child document nodes.

Sebobo:
I don't see the problem. The full text root nodes can be checked at runtime which is already happening I think. This covers their subparts.

Sebobo:
> I know you're still discussing whether this is even a viable solution or not, but since it is for my use case, I tested it in my project. It didn't work for the way my nodes are set up, because there's often a ContentCollection between the childNode that's not supposed to be indexed and the contentNode that is hidden. So I think you should probably allow both `Neos.Neos:Content` and `Neos.Neos:ContentCollection` in your while condition. Without it, it immediately exits or rather doesn't even enter the loop (at least in my setup).

You are correct. And as I said before I think only the check for the full text root should be done independent of any nodetype. I'll adjust the PR.

Sebobo:
I updated and retested the code. It now checks for the full text roots, so we are independent of the nodetypes.

But I found another issue related to this. When making a parent visible again, its child nodes are not added to the index again if the parent is not a full text root. This should be fixed in another PR imo.
As soon as the child nodes are edited again, their content reappears in the search.

PR #414, opened 2025-03-07
!!! FEATURE: Neos 9.0 compatibility

dlubitz:
This PR provides Neos 9.0 compatibility. This is a breaking change and requires a new major release.

The unit and functional tests have not been adapted. Due to the changes in the CR it's not a simple replacement in the test. Rather they should be rewritten as behat tests.

Depends on: 
* https://github.com/neos/content-repository-search/pull/51 
* https://github.com/Flowpack/Flowpack.ElasticSearch/pull/118


markusguenther (approved):
Like the rest of the related PRs, I have tested it with docs.neos.io, and it works fine :)
Thanks for digging that deep.

PR #427, opened 2026-04-29
BUGFIX: Skip nodes with non-existing nodetypes

Sebobo:
Before this change, the indexer crashed with an error as `isIndexable` required a valid NodeType instance, but `getNodeType` can return `null`.

Now the node with the unknown nodetype is skipped and the problem is logged with the raw name of the problematic nodetype.


dlubitz (approved):
Makes sense! Thank you

PR #429, opened 2026-07-14
BUGFIX: Allow indexing and cleanup of non-default CRs via CLI

Sebobo:
The other commands already had the option to provide a custom CR id, the build and cleanup commands were missing it.


dlubitz (approved):
Thank you!

